PR #144, opened 2013-10-07
Fixes #4004 - Remove historical and harmful code from rudder-upgrade

nperron:
Fixes #4004 - Remove historical and harmful code from rudder-upgrade

See http://www.rudder-project.org/redmine/issues/4004


peckpeck:
Commit modified

PR #235, opened 2014-02-25
Bug 4517/dev/correct vzps use

ncharles:


peckpeck:
Closing because the PR has been rejected

PR #238, opened 2014-02-27
Bug 4498/dev/vzps and promox 2.8

ncharles:
Note that this pull request is based on https://github.com/Normation/rudder-packages/pull/235 as it impacts exactly the same code


peckpeck:
@ncharles the matching ticket http://www.rudder-project.org/redmine/issues/4498 has no review status, is this PR still valid ?

amousset:
This has been merged in CFEngine in https://github.com/cfengine/core/pull/1948, and we only use releases that include the patch.

PR #274, opened 2014-03-22
Make a valid srpm

coredumb:
Sorry it comes a bit late...
Indeed it overrides some of your modifications... AIX support

All sourcexx: files should indeed be populated in SOURCES/ prior to run the build on the spec file.
Resulting standalone SRPM is fully functionnal and can be rebuilt anywhere.
Tested on EL6 and Fedora17

### DIFF
--- a/rudder-agent/SOURCES/perl-prepare.sh
+++ b/rudder-agent/SOURCES/perl-prepare.sh
@@ -32,7 +32,7 @@ if [ ! -f 'perl-prepare.sh' ]; then
     fi
 fi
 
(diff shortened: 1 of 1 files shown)


peckpeck:
Hello @coredumb 
We included all this in http://www.rudder-project.org/redmine/issues/7257 (for major version after rudder 3.1) and in http://www.rudder-project.org/redmine/issues/6770 (for minor releases)

I hope this match what you wanted to see included in.

peckpeck:
Aleady included in the packaging.

PR #320, opened 2014-05-20
Fixes #4872 - Add exclusion for ignore_interfaces.st if already existing during rudder-upgrade

nperron:
Fixes #4872 - Add exclusion for ignore_interfaces.st if already existing during rudder-upgrade

cf http://www.rudder-project.org/redmine/issues/4872


peckpeck:
This is not how we want to do it.

PR #474, opened 2014-09-03
Fixes #5421: Make /etc/default/rudder-jetty a config file in the packagi...

Kegeruneku:
...ng

Ticket: http://www.rudder-project.org/redmine/issues/5421

To be merged in 2.11


Kegeruneku:
Alrighty then.

PR #579, opened 2015-01-07
Fixes #6083: Do not run shlibdeps on ubuntu 10.04

Kegeruneku:
Ticket: http://www.rudder-project.org/redmine/issues/6093

To be merged in 3.0


ncharles:
oh, the commit message is wrong. It refers to http://www.rudder-project.org/redmine/issues/6093

PR #581, opened 2015-01-07
Bug 6095/int/6095 do not use fortify source during cfengine build

Kegeruneku:
Ticket: http://www.rudder-project.org/redmine/issues/6095

To be merged in 3.0


peckpeck:
That's why we need to be more careful

Kegeruneku:
Yup

PR #584, opened 2015-01-08
Fixes #6105: Remove old unwanted indexes from LDAP if necessary

Kegeruneku:
Ticket: http://www.rudder-project.org/redmine/issues/6105

To be merged in 2.10


jooooooon:
This PR should not have been merged before Normation/ldap-inventory#59 is. It should also be targeted on the 2.11 branch, like Normation/ldap-inventory#59 should be.

PR #586, opened 2015-01-08
Fixes #6112: Strip -Wp from CFLAGS in rudder-agent SPEC file

Kegeruneku:
Ticket: http://www.rudder-project.org/redmine/issues/6112

To be merged on 3.0


fanf:
Effectivelly, we don't want to have a "-Wp," without anything else...

PR #587, opened 2015-01-12
Fixes #6117: Remove FORTIFY_SOURCE hack and add CFEngine patch to be com...

Kegeruneku:
...patible with it

Ticket: http://www.rudder-project.org/redmine/issues/6117

To be merged in 3.0


peckpeck:
OK

PR #588, opened 2015-01-12
Fixes #6035: Do not block on yum during rudder-agent installation

Kegeruneku:
Ticket: http://www.rudder-project.org/redmine/issues/6035

To be merged in 2.10


peckpeck:
OK

PR #589, opened 2015-01-13
Fixes #6120: Add a missing 'then' in rudder-inventory-ldap postinstall s...

Kegeruneku:
...cripts

Ticket: http://www.rudder-project.org/redmine/issues/6120

To be merged in 2.10


peckpeck:
OK good

PR #590, opened 2015-01-13
Fixes #6119: Agent regenerates the list of package available to install several time per run, causing massive performance hit

peckpeck:
http://www.rudder-project.org/redmine/issues/6119


ncharles:
@peckpeck you forgot to use it in the Makefile

ncharles:
ha, sorry, my bad, they are automagically included

PR #591, opened 2015-01-14
Fixes #6092: Package manpages properly

Kegeruneku:
Ticket: http://www.rudder-project.org/redmine/issues/6092

To be merged in 2.10

### DIFF
--- a/rudder-agent/SOURCES/rudder-agent.sh
+++ b/rudder-agent/SOURCES/rudder-agent.sh
@@ -1,1 +1,13 @@
-export PATH=$PATH:/var/rudder/cfengine-community/bin
+# Add CFEngine binaries to the PATH
+PATH=${PATH}:/var/rudder/cfengine-community/bin
+export PATH
+
+# Add CFEngine manpages to the MANPATH
+if [ -z "${MANPATH}" ]
+then
+    MANPATH=/opt/rudder/share/man
+else
+    MANPATH=/opt/rudder/share/man:${MANPATH}
+fi
+
+export MANPATH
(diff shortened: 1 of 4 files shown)


rudder-agent/SPECS/rudder-agent.spec
```diff
@@ -441,7 +448,7 @@ if [ ${NB_COPIED_BINARIES} -gt 0 ];then echo "CFEngine binaries copied to workdi
 
 for i in cf-agent cf-promises cf-key cf-execd cf-serverd cf-monitord cf-runagent
 do
-  %{ruddervardir}/cfengine-community/bin/${i} -M | gzip > %{ruddervardir}/share/man/man8/${i}.8.gz
+  %{ruddervardir}/cfengine-community/bin/${i} -M | gzip > %{rudderdir}/share/man/man8/${i}.8.gz
```
peckpeck: I don't see why we execute this command at postinst time instead of packaging time.
It would allow to keep the package clean with no binary alteration (think debsums).

peckpeck:
@Kegeruneku ping

Kegeruneku:
Well, let's try

rudder-agent/SOURCES/rudder-agent.sh
```diff
@@ -1 +1,10 @@
-export PATH=$PATH:/var/rudder/cfengine-community/bin
+# Add CFEngine binaries to the PATH
+export PATH=${PATH}:/var/rudder/cfengine-community/bin
```
peckpeck: This is not posix. You should write 
VAR=value
export VAR

peckpeck:
Very good !
Please update rudder-agent.sh and self merge

PR #592, opened 2015-01-15
Fixes #6132: Replace the old patch to handle 4994+ chars long lists in C...

Kegeruneku:
...FEngine correctly

Ticket: https://www.rudder-project.org/redmine/issues/6132

To be merged in 2.11


peckpeck:
Good, a lot less code :)

PR #593, opened 2015-01-15
Fixes #6083: Do not run shlibdeps on ubuntu 10.04

Kegeruneku:
Ticket: http://www.rudder-project.org/redmine/issues/6134

To be merged in 2.11

### DIFF
--- a/rudder-agent/SOURCES/patches/UBUNTU_10_04/0001-use-bundled-openssl.patch
+++ b/rudder-agent/SOURCES/patches/UBUNTU_10_04/0001-use-bundled-openssl.patch
@@ -11,9 +11,9 @@ diff -Naur debian/control debian/control
  Homepage: http://www.rudder-project.org
  
 diff -Naur debian/rules debian/rules
---- debian/rules	2014-10-14 19:25:24.254087132 +0200
-+++ debian/rules	2014-10-15 15:09:40.638623237 +0200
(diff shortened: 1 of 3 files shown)


rudder-agent/SOURCES/patches/UBUNTU_10_04/0001-use-bundled-openssl.patch
```diff
@@ -42,3 +42,12 @@ diff -Naur debian/rules debian/rules
  	dh_install --SOURCEDIR=$(CURDIR)/SOURCES/ rudder.conf /etc/ld.so.conf.d
  	# Install a verification script for cron
  	dh_install --SOURCEDIR=$(CURDIR)/SOURCES/ check-rudder-agent /opt/rudder/bin/
+@@ -107,7 +111,7 @@
+ #	dh_perl
+ 	dh_makeshlibs
+ 	dh_installdeb
+-	dh_shlibdeps
++#	dh_shlibdeps
```
peckpeck: This should work for all dpkg build.
We must try to avoid specific cases as much as possible. 
Can you do this on the original file instead ?

peckpeck:
@Kegeruneku ping

Kegeruneku:
Updated

peckpeck:
Good for me, merging

PR #605, opened 2015-01-28
Fixes #6196: properly check if ldap is up and running to avoid skipping ...

ncharles:
...ldap upgrade scripts

### DIFF
--- a/rudder-webapp/SOURCES/rudder-upgrade
+++ b/rudder-webapp/SOURCES/rudder-upgrade
@@ -142,8 +142,6 @@ LDAPSEARCH="ldapsearch ${LDAP_PARAMETERS} -LLL"
 LDAPMODRDN="ldapmodrdn ${LDAP_PARAMETERS}"
 LDAPADD="ldapadd ${LDAP_PARAMETERS}"
 
-LDAP_EXISTS=$(${LDAPSEARCH} -b "cn=rudder-configuration" -s base dn 2>/dev/null | grep -c "dn: cn=rudder-configuration" || true)
-
 PSQL="psql -q -h ${SQL_SERVER} -p ${SQL_PORT} -U ${SQL_USER}"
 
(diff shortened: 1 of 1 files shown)


ncharles:
@peckpeck @Kegeruneku : could you both have a look at that ?

rudder-webapp/SOURCES/rudder-upgrade
```diff
@@ -292,16 +290,20 @@ if [ -d ${CONFIGURATION_REPOSITORY}/.git ];then
 fi
 
 # Check the LDAP and PostgreSQL daemons status before upgrading
-if [ ${LDAP_EXISTS} -ne 0 ]; then
-  echo -n "INFO: Checking LDAP service status..."
-  retry_wrapper "${LDAPSEARCH} -s one -b 'cn=rudder-configuration' cn >/dev/null 2>&1" "LDAP"
-  echo " OK"
-fi
+
+# Using retry_wrapper is necessary because during an upgrade, slapd may be restarted, so we need to wait
```
jooooooon: I suggest using this instead, which is more generic:

```
${LDAPSEARCH} -x -b "" -s base subschemaSubentry
```

rudder-webapp/SOURCES/rudder-upgrade
```diff
@@ -292,16 +290,20 @@ if [ -d ${CONFIGURATION_REPOSITORY}/.git ];then
...
+# for it to be available again
+echo -n "INFO: Checking LDAP service status..."
+retry_wrapper "${LDAPSEARCH} -b cn=Subschema -s base '(objectClass=subschema)' dn >/dev/null 2>&1" "LDAP"
```
jooooooon: I suggest using this instead, which is more generic:

```
${LDAPSEARCH} -x -b "" -s base subschemaSubentry
```

peckpeck:
Good !

PR #620, opened 2015-02-10
Fixes #6016: Do not install manpages during post installation

Kegeruneku:
Ticket: http://www.rudder-project.org/redmine/issues/6016

To be merged in 2.11


peckpeck:
Good !

PR #622, opened 2015-02-23
Fixes #6286: Append the epoch to the ncf explicit dependency on rudder-t...

Kegeruneku:
...echniques SPEC to prevent miscomparison of version on RPM based systems

Ticket: http://www.rudder-project.org/redmine/issues/6286

To be merged in 2.10

### DIFF
--- a/rudder-techniques/SPECS/rudder-techniques.spec
+++ b/rudder-techniques/SPECS/rudder-techniques.spec
@@ -51,7 +51,7 @@ BuildArch: noarch
 #BuildRequires: gcc
 
 # Add Requires here - order is important
-Requires: ncf >= 0.201501221640
+Requires: ncf >= 1398866025:0.201501221640
 
 # The following lines allow this package (rudder-techniques) to 
 # replace the older rudder-policy-templates package.


jooooooon:
This seems wrong. The ncf package should not have an epoch prefixed to it at all. Maybe check that this isn't a package building bug ?

ncharles:
oh, there already is an epoch. If we remove it, the comparision won't work (there is no way 0.201XXXXXX can be larger than 1398866025 )

Kegeruneku:
@jooooooon The comparison, on yum and zypper, fail to compare versions with epoch, (assuming the epoch is 1 I guess), so we need to compare with versions with the epoch in it on them. It is not wrong, it is our fault for having used epochs in the first place :)

Kegeruneku:
Basically:
```
zypper vcmp 1398866025:0.2015 0.2016
1398866025:0.2015 is newer than 0.2016

zypper vcmp 1398866025:0.2015 1398866025:0.2016
1398866025:0.2015 is older than 1398866025:0.2016
```

jooooooon:
Yeah. I do understand how epochs work, thanks ;-) 

My point was that the ncf package should have an epoch of 0 or 1, because we created it after realising that we should not use epochs like we used to. However, reading through the building code, this doesn't seem to be the case. I guess we missed a trick there :-( 

So yeah, go ahead.

peckpeck:
OK, so let's go ahead. Sorry we'll have to keep it for a long time, except if we change the product name.

PR #624, opened 2015-03-04
Fixes #6337: Run dh_shlibdeps with the right LD_LIBRARY_PATH to prevent ...

Kegeruneku:
...foreign libraries from beeing ignored

Ticket: http://www.rudder-project.org/redmine/issues/6337

To be merged in 2.11


rudder-agent/debian/rules
```diff
@@ -112,7 +112,7 @@ binary-arch: install
 #	dh_perl
 	dh_makeshlibs
 	dh_installdeb
-	dh_shlibdeps -- --ignore-missing-info
+	LD_LIBRARY_PATH="$(CURDIR)/debian/tmp/opt/rudder/lib:$${LD_LIBRARY_PATH}" dh_shlibdeps -- --ignore-missing-info
```
peckpeck: Can you add a comment here to say why and for which distro we did that, to easily cleanup or spot future problems ?
Thanks

Kegeruneku:
PR updated.

peckpeck:
OK Good !

PR #628, opened 2015-03-10
Fixes #6380: Correct a macro expansion on rudder-agent's SPECfile that m...

Kegeruneku:
...ade AIX builds fail

Ticket: http://www.rudder-project.org/redmine/issues/6380

To be merged in 2.10


peckpeck:
OK, it works

PR #630, opened 2015-03-12
Fixes #6396: Using parsestringarrayidx on AIX leads to agent hanging up or promises validation errors (only on 2.11)

peckpeck:
http://www.rudder-project.org/redmine/issues/6396


Kegeruneku:
Ok good.

PR #631, opened 2015-03-12
Fixes #6399: Correct ncf-api-virtualenv conffiles on Debian packages

Kegeruneku:
Ticket: http://www.rudder-project.org/redmine/issues/6399

To be merged in 2.11


ncf-api-virtualenv/debian/dirs
```diff
@@ -1,2 +1,2 @@
 usr/share/ncf-api-virtualenv
-etc/apache2/conf.d
```
peckpeck: Why do you want to remove it ? Since you will create a link inside this directory.

peckpeck:
OK good

PR #634, opened 2015-04-08
Fixes #6477: Create a script to sign files using openssl on unix

peckpeck:
PR replaced by https://github.com/Normation/rudder-packages/pull/644

### DIFF
--- a/rudder-agent/SOURCES/signature.sh
+++ b/rudder-agent/SOURCES/signature.sh
@@ -1,0 +1,41 @@
+#!/bin/sh
+
+# To allow Rudder to provide its own version of openssl
+export PATH=/opt/rudder/bin:$PATH
+
+if ! type openssl 2>/dev/null >/dev/null
+then
+  true
+else
+  echo "ERROR: openssl binary is missing !"
+  exit 1
+fi
+
+# md4 md5 sha sha1 sha224 sha256 sha384 sha512 whirlpool
+# The oldest openssl we support is 0.9.8 and it supports sha512
+HASH=sha512
+
+# the file to sign
+FILE="$1"
+if [ ! -e "${FILE}" ]
+then
+  echo "ERROR: Cannot sign: The file ${FILE} doesn't exist"
+  exit 2
+fi
+
+# the key to use for signature
+PRIVKEY=./localhost.priv
+
+# cfengine  passphrase
+PASSPHRASE="Cfengine passphrase"
+
+# Create signature
+SIGNATURE=$(openssl dgst -passin "pass:${PASSPHRASE}" -${HASH} -hex -sign "${PRIVKEY}" < "${FILE}" | sed -e 's/.*= //')
+
+# Create a signature FILE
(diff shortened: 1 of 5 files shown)


rudder-agent/SOURCES/signature.sh
```diff
@@ -0,0 +1,41 @@
...
+
+# md4 md5 sha sha1 sha224 sha256 sha384 sha512 whirlpool
+# The oldest openssl we support is 0.9.8 and it supports sha512
+HASH=sha512
+
+# the file to sign
+FILE="$1"
+if [ ! -e "${FILE}" ]
+then
+  echo "Cannot sign: The file ${FILE} doesn't exist"
```
Kegeruneku: Missing an "ERROR: " prefix here

rudder-agent/SOURCES/signature.sh
```diff
@@ -0,0 +1,41 @@
+#!/bin/sh
+
+# To allow Rudder to provide its own version of openssl
+export PATH=/opt/rudder/bin:$PATH
+
+if openssl -h 2> /dev/null
```
Kegeruneku: "if type openssl" maybe instead ?

Kegeruneku:
I'd rather name this utility signature and not signature.sh :) the suffix is not desirable here nor needed !

rudder-agent/SOURCES/signature.sh
```diff
@@ -0,0 +1,41 @@
...
+then
+  true
+else
+  echo "ERROR: openssl binary is missing !"
+  exit 1
+fi
```
Kegeruneku: Why not use:
```
if ! type openssl 2> /dev/null
then
  echo "ERROR: openssl binary is missing !"
  exit 1
fi
```

You'd avoid a useless "true"

rudder-agent/SOURCES/signature.sh
```diff
@@ -0,0 +1,41 @@
...
+then
+  echo "Cannot sign: The file ${FILE} doesn't exist"
+  exit 2
+fi
+
+# the key to use for signature
+PRIVKEY=/var/rudder/cfengine-community/ppkeys/localhost.priv
+
+# cfengine  passphrase
+PASSPHRASE="Cfengine passphrase"
```
Kegeruneku: ...
facepalm

peckpeck:
There will be a signature.cmd for windows too, the .sh was here to explicitly distinguish which one will be for wich environment

peckpeck:
PR updated

Kegeruneku:
That's a good reason !

Kegeruneku:
Would you be kind enough to retarget this PR for 3.1 ?

Kegeruneku:
Also this script should be able to try to use Rudder's bundled OpenSSL version if no system one is available (/opt/rudder/bin/openssl)

peckpeck:
PR replaced by https://github.com/Normation/rudder-packages/pull/644

PR #635, opened 2015-04-08
Fixes #6476: Better rudder-server-root init script

Kegeruneku:
Ticket: http://www.rudder-project.org/redmine/issues/6476

To be merged in master

### DIFF
--- a/rudder-server-root/SOURCES/rudder-server-root.init
+++ b/rudder-server-root/SOURCES/rudder-server-root.init
@@ -10,11 +10,11 @@
 # Required-Stop:	$local_fs $remote_fs $network $syslog
 # Default-Start:	2 3 4 5
 # Default-Stop:		0 1 6
-# Short-Description:	Rudder root server
-# Description:		Rudder root server including all components required to run
+# Short-Description:	Rudder server
+# Description:		Rudder server including all components required to run
 ### END INIT INFO
 #
-# Copyright (C) 2011 Normation
+# Copyright (C) 2011-2015 Normation
 #
 # This program is free software; you can redistribute it and/or
 # modify it under the terms of the GNU General Public License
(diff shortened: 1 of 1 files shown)


Kegeruneku:
Note: I removed apache handling in this script, as I feel it is not our job to manipulate Apache actually (PostgreSQL is also not really in our scope, but we need it to be running, it is a core component enough to be handled by us).

peckpeck:
Shouldn't we change the scrip name too ?

rudder-server-root/SOURCES/rudder-server-root.init
```diff
@@ -32,67 +32,188 @@
...
+### Check PostgreSQL's status if we are rudder-db
+[ -e /opt/rudder/etc/server-roles.d/rudder-db ] && HAVE_POSTGRESQL=1
+
+### Check Jetty's status if we are rudder-webapp or rudder-inventory-endpoint
+[ -e /opt/rudder/etc/server-roles.d/rudder-inventory-endpoint ] && HAVE_JETTY=1
+[ -e /opt/rudder/etc/server-roles.d/rudder-webapp ] && HAVE_JETTY=1
+
+## Service names
+APACHE2=apache2
+if [ ! -x /etc/debian_version ] && [ ! -x /etc/SuSE-release ]; then
```
peckpeck: Isn't there a distribution independent methods to detect this ?

Kegeruneku:
I think we should move this script to rudder-agent, and call it "rudder", but this is largely out of scope of the current ticket :)

rudder-server-root/SOURCES/rudder-server-root.init
Kegeruneku: Not really. We used to check for the init.d script, but it would not work with non-sysvinit systems...

Do you have a suggestion for this ?

rudder-server-root/SOURCES/rudder-server-root.init
```diff
@@ -32,67 +32,188 @@
...
+
+service_action()
+{
+# Arguments:
+## Service name
+## Action to trigger (start/stop/...)
+## Is the role present on this machine (0/1)
+
+# If the role is present, run the action
+if [ ${3} -eq 1 ]; then
```
peckpeck: Use "${3}" to have an understandable message if there is no $3

rudder-server-root/SOURCES/rudder-server-root.init
```diff
@@ -32,67 +32,188 @@
...
+
+service_details()
+{
+# Arguments:
+## Service human-friendly name
+## Service name
+## Is the role present on this machine (0/1)
+
+${ECHO_BIN} -n "\n * ${BLUE}${1}${NORMAL}:"
+if [ ${3} -eq 1 ];then
```
peckpeck: Idem

rudder-server-root/SOURCES/rudder-server-root.init
```diff
@@ -32,67 +32,188 @@
...
+service_status()
+{
+# Arguments:
+## Service human-friendly name
+## Service name
+## String to search for in the service status as an "OK" condition
+## Is the role present on this machine (0/1)
+
+${ECHO_BIN} -n " * ${1}\t"
+if [ ${4} -ne 1 ];then
```
peckpeck: Idem

rudder-server-root/SOURCES/rudder-server-root.init
peckpeck: Maybe by checking the binary /usr/sbin/apache2 vs /usr/sbin/httpd

peckpeck: Why have it at all if you don't manage apache ?

Kegeruneku:
PR replaced by #637

PR #636, opened 2015-04-22
Fixes #6516: Add OpenSSL binary dependency on Debian rudder agents

Kegeruneku:
Ticket: http://www.rudder-project.org/redmine/issues/6516

To be merged in master


peckpeck:
OK good

PR #638, opened 2015-04-22
Fixes #6517: Add initial SELinux support

Kegeruneku:
Ticket: http://www.rudder-project.org/redmine/issues/6517

To be merged in master


Kegeruneku:
This PR uses setsebool, which is not the proper way (using a selinux property instead would be better.)

peckpeck:
@Kegeruneku you should retarget to 3.1

Kegeruneku:
Replaced by #639

PR #640, opened 2015-04-27
Fixes #6513: Add the new rudder init script

Kegeruneku:
Ticket: http://www.rudder-project.org/redmine/issues/6513

To be merged in 3.1

Also, PR of the new init script has already been done in #635


rudder-agent/SPECS/rudder-agent.spec
```diff
@@ -474,10 +476,12 @@ then
   /usr/sbin/mkitab "rudder-agent:23456789:once:/usr/bin/startsrc -s rudder-agent"
   # No need to tell init to re-read /etc/inittab, it does it automatically every 60 seconds
 %else
-  chkconfig --add rudder-agent
+  chkconfig --del rudder-agent
```
peckpeck: Doesn't this produce an error if there is no rudder-agent yet ?

rudder-agent/debian/rules
```diff
@@ -83,7 +83,8 @@ binary-arch: install
 #	dh_installlogrotate
 #	dh_installmime
 #	dh_python
-	dh_installinit --no-start -R # restart after upgrade (instead of stop, upgrade, start)
+	dh_installinit --update-rcd-params=disable --no-start -R # restart after upgrade (instead of stop, upgrade, start)
+	dh_installinit --name=rudder --no-start -R # restart after upgrade (instead of stop, upgrade, start)
```
peckpeck: Isn't the -R option a possible source of problem during upgrade ? From the doc : 

           Do not stop the init script until after the package upgrade has
           been completed. This is different than the default behavior, which
           stops the script in the prerm, and starts it again in the postinst.

rudder-jetty/SPECS/rudder-jetty.spec
```diff
@@ -171,9 +171,9 @@ fi
 if [ $1 -eq 1 ]
 then
 	# Set rudder-agent as service
-	chkconfig --add rudder-jetty
+	chkconfig --del rudder-jetty
 	%if 0%{?rhel} && 0%{?rhel} >= 6
-	chkconfig rudder-jetty on
+	chkconfig rudder-jetty off
```
peckpeck: Will off work after having called --del on it ?

rudder-server-root/debian/rules
```diff
@@ -50,7 +50,6 @@ binary-arch: install
 #	dh_installdocs
 #	dh_installexamples
 	dh_install --SOURCEDIR=$(CURDIR)/SOURCES/ rudder-server-root /opt/rudder/etc/server-roles.d/
-	dh_installinit --noscripts --no-start --no-restart-on-upgrade --name=rudder-server-root
```
peckpeck: Shouldn't you remove the old one here as well ?

rudder-agent/SOURCES/rudder.init
```diff
@@ -0,0 +1,219 @@
...
+### Check PostgreSQL's status if we are rudder-db
+[ -e /opt/rudder/etc/server-roles.d/rudder-db ] && HAVE_POSTGRESQL=1
+
+### Check Jetty's status if we are rudder-webapp or rudder-inventory-endpoint
+[ -e /opt/rudder/etc/server-roles.d/rudder-inventory-endpoint ] && HAVE_JETTY=1
+[ -e /opt/rudder/etc/server-roles.d/rudder-webapp ] && HAVE_JETTY=1
+
+## Service names
+APACHE2=apache2
+if [ ! -x /etc/debian_version ] && [ ! -x /etc/SuSE-release ]; then
```
peckpeck: I think [ -d /etc/apache2 ] would be a better alternative since we base the decision on apache presence.

rudder-agent/SOURCES/rudder.init
```diff
@@ -0,0 +1,219 @@
...
+service_action()
+{
+# Arguments:
+## Service name
+## Action to trigger (start/stop/...)
+## Is the role present on this machine (0/1)
+
+# If the role is present, run the action
+if [ "${3}" -eq 1 ]; then
+  service ${1} ${2} || RETURN=${?}
```
peckpeck: As a precaution you should use "${var}" instead of ${var} when it's used a parameter

rudder-agent/SOURCES/rudder.init
peckpeck: Or just remove it since you don't manage apache in this script

rudder-agent/SOURCES/rudder.init
```diff
@@ -0,0 +1,219 @@
...
+# GNU General Public License for more details.
+#
+# GPL License: http://www.gnu.org/licenses/gpl.txt
+#
+#====================================================================
+
+# Variables
+
+## Colors configuration
+ECHO_BIN="/bin/echo -e"
```
peckpeck: This won't work on some os like AIX
You may use printf instead, which is defined in posix, but then you have to append \n at the end of each of your strings.

rudder-agent/SOURCES/rudder.init
```diff
@@ -0,0 +1,219 @@
...
+fi
+}
+
+start_services()
+{
+  service_action ${RUDDER_SLAPD} start ${HAVE_SLAPD}
+  service_action ${POSTGRESQL} start ${HAVE_POSTGRESQL}
+  service_action ${RUDDER_JETTY} start ${HAVE_JETTY}
+  service_action ${RUDDER_AGENT} start ${HAVE_AGENT}
+}
```
peckpeck: Couldn't we have an aggregated exit code for start and stop too ?

peckpeck:
Aside of those comments, this is good.

rudder-agent/debian/rules
Kegeruneku: We avoid at all costs starting/stopping scripts during install and upgrades, to be consistent between all our packages (and prevent os-dependent surprises). Post installation / migration scripts and the agent take care of restarting what needs restarting when needed !

rudder-jetty/SPECS/rudder-jetty.spec
Kegeruneku: Looks so:
```
[root@centos-6-64 ~]# chkconfig --del rudder-agent && echo $?
0
[root@centos-6-64 ~]# chkconfig rudder-agent off && echo $?
0
```

rudder-agent/SOURCES/rudder.init
Kegeruneku: Yup.

rudder-agent/SOURCES/rudder.init
Kegeruneku: OK

rudder-agent/SOURCES/rudder.init
Kegeruneku: OK

Kegeruneku: (Just as a note: We do not use this script on AIX, SRC takes care of the init stuff instead :) )

rudder-agent/SOURCES/rudder.init
Kegeruneku: OK

rudder-agent/SPECS/rudder-agent.spec
Kegeruneku: Indeed.

rudder-server-root/debian/rules
Kegeruneku: Nope, I will not remove the rudder-server-root role file :) it would cause a lot of problems...

rudder-agent/SOURCES/rudder.init
peckpeck: Not only on AIX, but on systems with non GNU echo.

rudder-agent/debian/rules
peckpeck: So the server will still be running even if we upgrade the database structure ?

rudder-jetty/SPECS/rudder-jetty.spec
peckpeck: OK

rudder-server-root/debian/rules
peckpeck: Not the role file, the init file

rudder-agent/debian/rules
Kegeruneku: Yes, since the database migration done by rudder-upgrade will trigger a PostgreSQL restart

Kegeruneku: (And a Jetty one)

rudder-server-root/debian/rules
Kegeruneku: Oh you mean git rm ?

rudder-agent/debian/rules
peckpeck: ok

rudder-server-root/debian/rules
peckpeck: No I mean dh_installinit --update-rcd-params=disable --no-start -R

rudder-agent/SOURCES/rudder.init
```diff
@@ -0,0 +1,241 @@
...
+{
+# Arguments:
+## Service name
+## Action to trigger (start/stop/...)
+## Is the role present on this machine (0/1)
+
+# If the role is present, run the action
+if [ "${3}" -eq 1 ]; then
+  printf " * ${1}...\t"
+  service "${1}" "${2}" >/dev/null 2>&1 || RETURN="${?}"
```
peckpeck: Here RETURN is a global variable without initial value.
So it will always be empty or the content of the last error, so if there is an error on a service, the next service will be said to have an error too

rudder-agent/SOURCES/rudder.init
```diff
@@ -0,0 +1,241 @@
...
+
+  if [ "${RETURN}" -ne 0 ]; then
```
peckpeck: I think you should just use $? here

rudder-agent/SOURCES/rudder.init
```diff
@@ -0,0 +1,241 @@
...
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# GPL License: http://www.gnu.org/licenses/gpl.txt
+#
+#====================================================================
+
+# Utility functions
+
+get_service_name()
```
peckpeck: There could be a comment indicating what this function intends to do, and especially that $1 is a grep pattern

rudder-agent/SOURCES/rudder.init
peckpeck: OK, ignore this, it's only the $? below that should be changed

rudder-agent/SOURCES/rudder.init
```diff
@@ -0,0 +1,241 @@
...
+  service_action "${POSTGRESQL}"   stop "${HAVE_POSTGRESQL}"
+  service_action "${RUDDER_JETTY}" stop "${HAVE_JETTY}"
+  service_action "${RUDDER_AGENT}" stop "${HAVE_AGENT}"
+}
+
+details_services()
+{
+  printf "${GREEN}Detailed component status:${NORMAL}\n"
+
+  RETURN=0
```
peckpeck: There's already a default above

rudder-agent/SOURCES/rudder.init
```diff
@@ -0,0 +1,241 @@
...
+  service_details "LDAP DB"     "${RUDDER_SLAPD}" "${HAVE_SLAPD}"
+  service_details "SQL DB"      "${POSTGRESQL}"   "${HAVE_POSTGRESQL}"
+  service_details "Application" "${RUDDER_JETTY}" "${HAVE_JETTY}"
+}
+
+status_services()
+{
+  printf "${GREEN}Component status:${NORMAL}\n"
+
+  RETURN=0
```
peckpeck: There's already a default above

peckpeck:
Ok, I find it in a good shape

rudder-server-root/debian/rules
peckpeck: or update-rc.d --force remove ...

peckpeck:
OK Good

PR #632, opened 2015-03-21
Fix SLES SP identification

matya:
Fixed missing "cat" in detect_os.sh for SLES SP's

PR #642, opened 2015-04-27
Fixes #6486: Correct SLES Service Pack identification in detect_os.sh

Kegeruneku:
Fixed missing "cat" in detect_os.sh for SLES SP's

Ticket:  http://www.rudder-project.org/redmine/issues/6486

This PR is only a retarget of #632


Kegeruneku:
PR has already been reviewed and accepted in #632, self merging.

Kegeruneku:
Thanks a lot, @matya !!!

PR #643, opened 2015-04-27
Fixes #2882: Package a SELinux policy for rudder-webapp

Kegeruneku:
Ticket: http://www.rudder-project.org/redmine/issues/2882

To be merged in 3.1

### DIFF
--- a/rudder-webapp/debian/dirs
+++ b/rudder-webapp/debian/dirs
@@ -1,6 +1,8 @@
 opt/rudder/bin
 opt/rudder/share/webapps
 opt/rudder/share/rudder-plugins
+opt/rudder/share/certificates
+opt/rudder/share/selinux
 opt/rudder/etc
 opt/rudder/etc/ssl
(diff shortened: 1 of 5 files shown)


rudder-webapp/debian/dirs
```diff
@@ -1,6 +1,8 @@
 opt/rudder/bin
 opt/rudder/share/webapps
 opt/rudder/share/rudder-plugins
+opt/rudder/share/certificates
```
peckpeck: What is this directory doing here ?

peckpeck:
OK, this is a good PR

rudder-webapp/debian/dirs
Kegeruneku: It has been added to be consistent with the SPEC file (and it corrects a small mistake, this dir belongs to the rudder-webapp package and had been forgotten when we added cert management)

peckpeck:
OK Good

PR #646, opened 2015-04-29
Fixes #6547: Add SLES12 support

Kegeruneku:
Ticket: http://www.rudder-project.org/redmine/issues/6547

To be merged in 3.1

### DIFF
--- a/ncf/SOURCES/.dependencies
+++ b/ncf/SOURCES/.dependencies
@@ -1,2 +1,0 @@
-SLES11:gcc
-SLES10:gcc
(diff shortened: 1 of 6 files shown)


ncf/SPECS/ncf.spec
```diff
@@ -49,8 +49,6 @@ Source1: rudder-sources
 BuildRoot: %{_tmppath}/%{name}-%{version}-%{release}-root-%(%{__id_u} -n)
 BuildArch: noarch
 
-#BuildRequires: gcc
```
Kegeruneku: Seriously, let's clean that up. This package only drops a folder on the target machine :)

ncf/SOURCES/.dependencies
```diff
@@ -1,2 +0,0 @@
-SLES11:gcc
-SLES10:gcc
```
Kegeruneku: Lol nope.

peckpeck:
OK Good

PR #649, opened 2015-05-04
Fixes #6551: signature.sh doesn't use absolute key path

peckpeck:
http://www.rudder-project.org/redmine/issues/6551


amousset:
Shouldn't this be in 3.1 branch too?

PR #650, opened 2015-05-04
Fixes #6549: patches does not apply anymore on rudder-agent

VinceMacBuche:
http://www.rudder-project.org/redmine/issues/6549


peckpeck:
OK Good

PR #651, opened 2015-05-04
Fixes #6555: sles_version macro is no more available on SLE12, preventing agent build

VinceMacBuche:
http://www.rudder-project.org/redmine/issues/6555


peckpeck:
OK Good

PR #653, opened 2015-05-06
Fixes #6567: Add a script to manage node keys on server

peckpeck:
http://www.rudder-project.org/redmine/issues/6567

### DIFF
--- a/rudder-webapp/SOURCES/rudder-keys
+++ b/rudder-webapp/SOURCES/rudder-keys
@@ -1,0 +1,90 @@
+#!/bin/sh
+#####################################################################################
+# Copyright 2015 Normation SAS
+#####################################################################################
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, Version 3.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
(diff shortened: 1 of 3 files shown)


peckpeck:
PR updated

amousset:
Should this be added to rudder cli ? (like rudder server debug).

rudder-webapp/SOURCES/rudder-keys
```diff
@@ -0,0 +1,90 @@
...
+
+# Commands
+LDAP_PARAMETERS="-H ldap://${LDAP_SERVER}:${LDAP_PORT}/ -D ${LDAP_USER} -w ${LDAP_PASSWORD} -x"
+
+
+if [ "${ACTION}" = "reset-status" ]
+then
+
+  ldapmodify ${LDAP_PARAMETERS} <<EOF
+dn: nodeId=nodeId=${UUID},${ACCEPTED_NODES_BASE}
```
amousset: There are two "nodeId=" in the dn.

peckpeck: Oops I thought I had corrected it

peckpeck:
I don't think this should be added to rudder cli.
This is a temporary script for a feature that should be added to the interface and to the API. It will then be time to remove the script.

VinceMacBuche:
After discussing with @fanf  and @peckpeck, we agreed on the fact that those functions should not be in rudder-cli, and that they would be integrated via api functions (and rudder-api-cli) 

However we will upgrade the error messages in ldap-inventory to talk about that script

VinceMacBuche:
Merging!

VinceMacBuche:
Not merging ;) the dn 'nodeid=nodeid=' is not fixed @peckpeck  ;)

peckpeck:
PR updated

VinceMacBuche:
Thanks @peckpeck ! merging

PR #654, opened 2015-05-12
Fixes #6530: Cannot save technique with ncf builder - error during the API call to the technique librairy reload

peckpeck:
http://www.rudder-project.org/redmine/issues/6530


ncharles:
Thank you !

PR #656, opened 2015-05-15
Fixes #6587: cfengine context is not updated properly when doing multiple indivudual package actions

amousset:
http://www.rudder-project.org/redmine/issues/6587


ncharles:
oh, is it mandatory to have one patch per branch ?

PR #657, opened 2015-05-15
Fixes #6590:  cfengine context is not updated properly when doing multiple indivudual package actions for 3.0

amousset:
http://www.rudder-project.org/redmine/issues/6590


ncharles:
awesomeness overflow !

ncharles:
@amousset you should target branch 2.11, as the problem exists also in this branch

ncharles:
see also https://github.com/Normation/rudder-packages/pull/656

@amousset is it necessary to have one patch per branch ?

peckpeck:
Yes since our naming and numbering of patches have changed between branches.

ncharles:
this numbering is quite stupid, it will always prevent direct merge between branches...

ncharles:
the commit is missing the actual application of patch, in the makefile

PR #659, opened 2015-05-15
Fixes #6597: Rudder webapp postinst are  not valid

VinceMacBuche:
http://www.rudder-project.org/redmine/issues/6597


peckpeck:
Ok Good

PR #661, opened 2015-05-18
Fixes #6605: remove duplicated patch on cfengine

amousset:
http://www.rudder-project.org/redmine/issues/6605


peckpeck:
OK Good

PR #663, opened 2015-05-19
Fixes #6617: ncf post-hooks fails when there is nothing to commit

VinceMacBuche:
http://www.rudder-project.org/redmine/issues/6617


peckpeck:
OK Good

PR #664, opened 2015-05-21
Fixes #6632: Install the init scripts we need without dh_installinit

Kegeruneku:
Ticket: http://www.rudder-project.org/redmine/issues/6632

To be merged in 3.1


peckpeck:
OK good

PR #665, opened 2015-05-21
Fixes #6652: Correct SELinux support for RHEL

Kegeruneku:
Ticket: http://www.rudder-project.org/redmine/issues/6652

To be merged in 3.1


peckpeck:
Thank you for the fix !

PR #666, opened 2015-05-22
Fixes #6651: Provide a fallback name in case a service is not found a…

Kegeruneku:
…nd remove 'skipped' messages in rudder init script

Ticket: http://www.rudder-project.org/redmine/issues/6651

To be merged in 3.1

### DIFF
--- a/rudder-agent/SOURCES/rudder.init
+++ b/rudder-agent/SOURCES/rudder.init
@@ -47,6 +47,11 @@ get_service_name()
     SERVICE_NAME=$(ls -1 /etc/init.d | grep "${1}" | tail -n 1)
   fi
 
+  # If we can not get the service name, fallback to the user-supplied name
+  if [ -z "${SERVICE_NAME}" ]; then
+    SERVICE_NAME="${1}"
+  fi
+
   echo ${SERVICE_NAME}
 }
 
@@ -69,8 +74,6 @@ if [ "${3}" -eq 1 ]; then
     printf "${GREEN}OK${NORMAL}\n"
   fi
 
-else
-  echo "* ${1}: Role is not present on this machine, skipped."
 fi
 }
 


rudder-agent/SOURCES/rudder.init
```diff
@@ -47,6 +47,11 @@ get_service_name()
     SERVICE_NAME=$(ls -1 /etc/init.d | grep "${1}" | tail -n 1)
   fi
 
+  # If we can not get the service name, fallback to the user-supplied name
```
Kegeruneku: This prevents the case where a script / service can not be found and returns an empty echo. We should at lease give back what the user proposed.

rudder-agent/SOURCES/rudder.init
```diff
@@ -69,8 +74,6 @@ if [ "${3}" -eq 1 ]; then
     printf "${GREEN}OK${NORMAL}\n"
   fi
 
-else
-  echo "* ${1}: Role is not present on this machine, skipped."
```
Kegeruneku: This is unneeded, and spams the user multiple times with the same message every time a user restarts the service:

Exemple on an agent only machine:
```
[root@centos-6-64 ~]# ./rudder.init restart
Stopping Rudder services:
 * rudder-slapd: Role is not present on this machine, skipped.
 * postgresql: Role is not present on this machine, skipped.
 * rudder-jetty: Role is not present on this machine, skipped.
 * rudder-agent...	OK

Starting Rudder services:
 * rudder-slapd: Role is not present on this machine, skipped.
 * postgresql: Role is not present on this machine, skipped.
 * rudder-jetty: Role is not present on this machine, skipped.
 * rudder-agent...	OK
```

This is better:
```
[root@centos-6-64 ~]# ./rudder.init restart
Stopping Rudder services:
 * rudder-agent...	OK

Starting Rudder services:
 * rudder-agent...	OK
```

peckpeck:
OK good

PR #667, opened 2015-05-25
Fixes #6669: Package rudder-inventory-ldap initi script are not correctly named and packaged

VinceMacBuche:
http://www.rudder-project.org/redmine/issues/6669


Kegeruneku:
OK, good.

PR #668, opened 2015-05-26
Fixes #6670: Package rudder-jetty init script are not correctly named and packaged

VinceMacBuche:
http://www.rudder-project.org/redmine/issues/6670


Kegeruneku:
Merged/

PR #669, opened 2015-05-26
Fixes #6672: Remove packaging of rudder-webapp.pp on debian

VinceMacBuche:
http://www.rudder-project.org/redmine/issues/6672

### DIFF
--- a/rudder-webapp/debian/rules
+++ b/rudder-webapp/debian/rules
@@ -120,9 +120,6 @@ binary-arch: install
 	dh_install --SOURCEDIR=$(CURDIR)/SOURCES/rudder-doc pdf /usr/share/doc/rudder
 	dh_install --SOURCEDIR=$(CURDIR)/SOURCES/rudder-doc html /usr/share/doc/rudder
 
-	# Install SELinux policies
-	dh_install --SOURCEDIR=$(CURDIR)/SOURCES/ rudder-webapp.pp /opt/rudder/share/selinux/
-
 	# Install rudder-keys
 	dh_install --SOURCEDIR=$(CURDIR)/SOURCES/ rudder-keys /opt/rudder/bin/
 


rudder-webapp/debian/rules
```diff
@@ -120,8 +120,8 @@ binary-arch: install
 	dh_install --SOURCEDIR=$(CURDIR)/SOURCES/rudder-doc pdf /usr/share/doc/rudder
 	dh_install --SOURCEDIR=$(CURDIR)/SOURCES/rudder-doc html /usr/share/doc/rudder
 
-	# Install SELinux policies
-	dh_install --SOURCEDIR=$(CURDIR)/SOURCES/ rudder-webapp.pp /opt/rudder/share/selinux/
+	# Install SELinux policies -> Not done yet, only on RHEL
+	# dh_install --SOURCEDIR=$(CURDIR)/SOURCES/ rudder-webapp.pp /opt/rudder/share/selinux/
```
Kegeruneku: We completely removed SELinux support from Debian packages at the moment. Please remove this instead of commenting it, no dead code :)

VinceMacBuche: Yeah I thought about that too.

But since one day we will add the selinux policy (even on debian), so i thought it would be great just to uncomment that line

Removing it then

Kegeruneku: Don't worry, it is trivial enough not to require a tremendous amount of thinking to add it again :)

VinceMacBuche:
PR updated

VinceMacBuche:
PR updated

rudder-webapp/debian/rules
VinceMacBuche: It's not about being trivial, it's about me being lazy ;)

Kegeruneku:
Accepted !

PR #670, opened 2015-05-27
Fixes #6679: Copy the rudder-webapp SELinux policy to the right locat…

Kegeruneku:
…ion before building it

Ticket: http://www.rudder-project.org/redmine/issues/6679

To be merged in 3.1


peckpeck:
OK Good

PR #672, opened 2015-05-28
Fixes #6681: rudder-webapp spec tries to copy rudder-webapp.pp from wrong directory

VinceMacBuche:
http://www.rudder-project.org/redmine/issues/6681


Kegeruneku:
Yes !§ thank you !

PR #673, opened 2015-05-29
Fixes #6682: rudder-webapp spec tries to build rudder-webapp.pp even if he can't

amousset:
http://www.rudder-project.org/redmine/issues/6682


Kegeruneku:
Accepted.

PR #674, opened 2015-06-03
Fixes #6263: Disable other init scripts in rudder-server-root to prev…

Kegeruneku:
…ent issues in startup ordering

Ticket: http://www.rudder-project.org/redmine/issues/6263

To be merged in 2.11

### DIFF
--- a/rudder-server-root/debian/postinst
+++ b/rudder-server-root/debian/postinst
@@ -20,6 +20,13 @@ set -e
 
 case "$1" in
     configure)
+
+        # On a rudder-root-server, we disable all rudder-* init scripts
+        # to prevent interference with rudder-server-root init.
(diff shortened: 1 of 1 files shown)


VinceMacBuche:
Ok great!

VinceMacBuche:
Hmm wait, when techniques needs to restart processes, do they use those scripts ?

Kegeruneku:
Yes. Disabling them in rc.d does not mean they are not usable any more of course :)

Kegeruneku:
Just that they will not be called during system startup anymore.

VinceMacBuche:
ha ok, it removed them from the startup not from the init scripts ;)

Kegeruneku:
Yes. The "remove" argument from update-rc.d is ... kinda misleading.

VinceMacBuche:
Merging then!

PR #675, opened 2015-06-03
Fixes #6697: Problem with MANPATH on rudder-agent 3.0.5 on RHEL/CentOS

VinceMacBuche:
http://www.rudder-project.org/redmine/issues/6697


Kegeruneku:
Accepted

PR #676, opened 2015-06-04
Fixes #6702: RHEL7 does not initialise postgresql db

Kegeruneku:
http://www.rudder-project.org/redmine/issues/6702


peckpeck:
OK easy one

PR #677, opened 2015-06-04
Fixes #6699: rudder-reports does not find correct postgresql service when using systemctl

Kegeruneku:
http://www.rudder-project.org/redmine/issues/6699

### DIFF
--- a/rudder-reports/SPECS/rudder-reports.spec
+++ b/rudder-reports/SPECS/rudder-reports.spec
@@ -109,7 +109,7 @@ install -m 644 %{SOURCE4} %{buildroot}/opt/rudder/etc/server-roles.d/
 #=================================================
 
 if type systemctl >/dev/null 2>&1; then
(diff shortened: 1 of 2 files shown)


rudder-reports/SPECS/rudder-reports.spec
```diff
@@ -109,7 +109,7 @@ install -m 644 %{SOURCE4} %{buildroot}/opt/rudder/etc/server-roles.d/
 #=================================================
 
 if type systemctl >/dev/null 2>&1; then
-  POSTGRESQL_SERVICE_NAME=$(systemctl -a | awk -F'.' '{print $1}' | grep "postgresql" | tail -n 1)
+  POSTGRESQL_SERVICE_NAME=$(systemctl list-unit-files --type service | awk -F'.' '{print $1}' | grep -E "^postgresql[0-9]*$" | tail -n 1)
```
Kegeruneku: VME gave me this command, it's much more effective :)

I modified the grep to match only main postgresql services and not sub services for each cluster (like postgresql@9xx...)

peckpeck:
OK Good

PR #684, opened 2015-06-15
Fixes #6724: Cannot save technique with ncf builder - error during th…

ncharles:
…e API call to the technique librairy reload


VinceMacBuche:
Merging, same as https://github.com/Normation/rudder-packages/pull/654 but in correct branch!

PR #696, opened 2015-06-29
Fixes #6848: Failed migration on first install

VinceMacBuche:
https://www.rudder-project.org/redmine/issues/6848


VinceMacBuche:
PR updated

Kegeruneku:
Please keep the old if :) else it would break the migration logic

VinceMacBuche:
PR updated

PR #709, opened 2015-07-15
Fixes #6990: Remove cf-clerk from rudder-packages

VinceMacBuche:
https://www.rudder-project.org/redmine/issues/6990


peckpeck:
OK Good

PR #710, opened 2015-07-16
Fixes #6996: If we ask to install a package in a version at least equal to another, we get no report if the version installed is > than the one requested

amousset:
https://www.rudder-project.org/redmine/issues/6996


ncharles:
yeah \o/

PR #711, opened 2015-07-16
Fixes #7002: If we ask to install a package in a version at least equal to another, we get no report if the version installed is > than the one requested (on 3.0)

amousset:
https://www.rudder-project.org/redmine/issues/7002


amousset:
Same patch for the 3.1 branch.

PR #721, opened 2015-07-29
Fixes #6754: Xen list parsed badly by FusionInventory, causing VCPU to be counted badly

Kegeruneku:
https://www.rudder-project.org/redmine/issues/6754


peckpeck:
OK Good

PR #723, opened 2015-07-30
Fixes #7048: The package need some adjustments on solaris 3.2

peckpeck:
https://www.rudder-project.org/redmine/issues/7048


Kegeruneku:
My eyes... :(

But acceptable !

PR #725, opened 2015-07-30
Fixes #7021: When SELinux is enabled, the ncf-api-venv home is owned …

amousset:
…by root

https://www.rudder-project.org/redmine/issues/7021

We can avoid the SELinux problem by creating the /var/lib/ncf-api-venv directory manually. It also avoids copying the skel content into it.


Kegeruneku:
OK!

PR #726, opened 2015-07-30
Fixes #7053: With SELinux enabled, httpd cannot execute ncf scripts

amousset:
https://www.rudder-project.org/redmine/issues/7053


peckpeck:
OK Good

PR #727, opened 2015-08-04
Fixes #7068: Inventories are not signed under solaris

peckpeck:
https://www.rudder-project.org/redmine/issues/7068


Kegeruneku:
OK !

PR #730, opened 2015-08-06
Fixes #7076: Broken SELinux packaging

amousset:
https://www.rudder-project.org/redmine/issues/7076


Kegeruneku:
Much better :dancer:

PR #731, opened 2015-08-06
Fixes #6127: Rsyslog problem on Ubuntu 14.04 fresh install

Kegeruneku:
https://www.rudder-project.org/redmine/issues/6127


peckpeck:
Ok Good

PR #736, opened 2015-08-12
Fixes #7097: Error starting apache on SLES11 relay

amousset:
https://www.rudder-project.org/redmine/issues/7097


Kegeruneku:
OK

PR #741, opened 2015-08-26
Fixes #7133: Create the migration script for the database to correct …

ncharles:
…the serialization of groups in the rules


ncharles:
to be merged at the same time as https://github.com/Normation/rudder/pull/919

Kegeruneku:
Looks perfect to me !

PR #743, opened 2015-09-04
Fixes #7161: Create a migration script for the database change for hi…

ncharles:
…storization of global agent schedule


ncharles:
to be merged at the same time as https://github.com/Normation/rudder/pull/922

Kegeruneku:
OK !

PR #748, opened 2015-09-21
Fixes #7213: Use tab to ident in debian/rules or else it fails

VinceMacBuche:
https://www.rudder-project.org/redmine/issues/7213


Kegeruneku:
I accept

PR #753, opened 2015-10-06
Fixes #7249: Invalid variable name for vzps command in check/init script of rudder-agent

VinceMacBuche:
https://www.rudder-project.org/redmine/issues/7249


peckpeck:
OK Good

PR #754, opened 2015-10-07
Fixes #7253: Patch adding Rudder.pm is noit correct

VinceMacBuche:
https://www.rudder-project.org/redmine/issues/7253


peckpeck:
OK Good

PR #759, opened 2015-10-12
Fixes #7272: Ubuntu 10.04 patches does not apply anymore

VinceMacBuche:
https://www.rudder-project.org/redmine/issues/7272


Kegeruneku:
Accepted

PR #760, opened 2015-10-12
Fixes #7273: if compilation flags tests fails, rpm build fails

VinceMacBuche:
https://www.rudder-project.org/redmine/issues/7273

### DIFF
--- a/rudder-agent/SPECS/rudder-agent.spec
+++ b/rudder-agent/SPECS/rudder-agent.spec
@@ -301,8 +301,7 @@ TRY_CFLAGS="-fPIE -fstack-protector"
 
 FLAG_TEST_FILE=`mktemp /tmp/hardening.XXXXXX`
 echo "void main() {}" > "${FLAG_TEST_FILE}.c"
-gcc ${TRY_CFLAGS} ${TRY_LDFLAGS} -o "${FLAG_TEST_FILE}" "${FLAG_TEST_FILE}.c" 2>/dev/null
-if [ $? -eq 0 ]
+if gcc ${TRY_CFLAGS} ${TRY_LDFLAGS} -o "${FLAG_TEST_FILE}" "${FLAG_TEST_FILE}.c" >/dev/null 2>&1
 then
   SECURE_CFLAGS="${TRY_CFLAGS}"
   SECURE_LDFLAGS="${TRY_LDFLAGS}"


rudder-agent/SPECS/rudder-agent.spec
```diff
@@ -301,8 +301,7 @@ TRY_CFLAGS="-fPIE -fstack-protector"
 
 FLAG_TEST_FILE=`mktemp /tmp/hardening.XXXXXX`
 echo "void main() {}" > "${FLAG_TEST_FILE}.c"
-gcc ${TRY_CFLAGS} ${TRY_LDFLAGS} -o "${FLAG_TEST_FILE}" "${FLAG_TEST_FILE}.c" 2>/dev/null
-if [ $? -eq 0 ]
+if [ $(gcc ${TRY_CFLAGS} ${TRY_LDFLAGS} -o "${FLAG_TEST_FILE}" "${FLAG_TEST_FILE}.c" >/dev/null 2>&1; echo $?) -eq 0 ]
```
peckpeck: I think the problem is set -e and this scheme will fail with set -e too.
Instead use: if gcc ..... ; then true; else ... ; fi

VinceMacBuche:
PR updated

rudder-agent/SPECS/rudder-agent.spec
VinceMacBuche: It does not fail in set -e 

I tried with the following script: 

     #!/bin/sh

      set -e

    SECURE_CFLAGS="NO FLAGS"
    TRY_LDFLAGS="-pie -Wl,-z,relro -Wl,-z,now"
    TRY_CFLAGS="-fPIE -fstack-protector"
    FLAG_TEST_FILE=`mktemp /tmp/hardening.XXXXXX`
    echo "void main() {}" > "${FLAG_TEST_FILE}.c"
    if gcc ${TRY_CFLAGS} ${TRY_LDFLAGS} -o "${FLAG_TEST_FILE}" "${FLAG_TEST_FILE}.c" >/dev/null 2>&1
     then
       SECURE_CFLAGS="${TRY_CFLAGS}"
       SECURE_LDFLAGS="${TRY_LDFLAGS}"
    fi
    echo $SECURE_CFLAGS

VinceMacBuche:
PR updated

peckpeck:
OK Good

PR #764, opened 2015-10-19
Fixes #7293: Temporary files are not cleaned up after build

peckpeck:
https://www.rudder-project.org/redmine/issues/7293


Kegeruneku:
Looks good :)

PR #766, opened 2015-10-23
Fixes #7301: promises are not regenerated when a node is transformed into a relay

VinceMacBuche:
https://www.rudder-project.org/redmine/issues/7301

### DIFF
--- a/rudder-server-root/SOURCES/rudder-node-to-relay
+++ b/rudder-server-root/SOURCES/rudder-node-to-relay
@@ -233,4 +233,8 @@ EOF
 
 ${ECHO} " Done."
 
+${ECHO} -n "INFO: Triggering promises generation..."
+curl -k "https://localhost/rudder/api/deploy/reload"
+${ECHO} " Done."
+
 ${ECHO} -e "\nSuccess: all steps completed successfully."


peckpeck:
Ok Good

matya:
@peckpeck  @VinceMacBuche, why not use http://localhost:8080/ directly? You could make sure you hit the rudder API... I am just asking because we for example modified rudder's vhost config to listen on https only on a dedicated Service IP instead of *:443, as there are also other services running on https on that host...

peckpeck:
@VinceMacBuche what do you think of @matya's proposal ?

VinceMacBuche:
I have never thought about using the jetty endpoint directly, but you are totally right @matya, vhost may be modified and only :8080 is always available ...

Modifying the PR !

jooooooon:
I don't think that is a good idea. The service architecture in Rudder is designed so that the *single* point of entry to the web applications is via an Apache server. This leaves us a lot of freedom to change jetty into another application server, or split out the two webapps (inventory-endpoint and rudder-web) into two independant application containers (as planned in http://www.rudder-project.org/redmine/issues/2630) - which would therefore possibly no longer be listening on localhost:8080.

There is nothing to enforce this currently, but I see no reason why we couldn't "lock down" application security by configuring jetty to only accept connections coming from apache, etc.

However, on a wider scale, if you consider the "split-server" installation architecture of a Rudder server, it would be good to have an internal communication path so that this script could be run anywhere, and we could hit the "Rudder API" URL just by relying on a DNS name (internal, hosts, or external).

Still, for this PR, I think going through Apache should be kept.

@matya, for your Apache config, could you also expose the Rudder API and webapp on localhost:443 (or localhost:80, I suppose)? I know this is kinda annoying to do in Apache config, just a thought...

VinceMacBuche:
I agree with @jooooooon  (in fact I thought i added a comment after my last one to state that, but it seems it get into a black hole ...)

VinceMacBuche:
Since it was ok (for @peckpeck and @jooooooon ) I am merging this, 

If you want to discuss this @matya, we can still continue here or on a new issue on our Redmine :)

Still many thanks about your feedbacks and opinions, they are very useful!

matya:
I can live with the workaround of binding additionally to 127.0.0.1:443 . 
However, as already mentioned, this script now is bound to the server hosting the web service.
It would be nicer if a promotion to a relay could be triggered directly via the REST API or made available as an administrative option in the WebGUI.

VinceMacBuche:
It is already bound to the server since it make modification to the ldap directly on localhost

And I Agrre with you it should be done via api and Web UI, I'm opening an issue about this in redmine :)

PR #767, opened 2015-10-27
Fixes #7315: Add migration scripts to authentication providers configuration

fanf:
https://www.rudder-project.org/redmine/issues/7315

### DIFF
--- a/rudder-webapp/SOURCES/rudder-upgrade
+++ b/rudder-webapp/SOURCES/rudder-upgrade
@@ -57,6 +57,7 @@ trap anomaly_handler ERR INT TERM
 # - 3.0.0        : Replace varchar by text in the database
 # - 3.0.2        : Check for the PostgreSQL version (>= 8.4)
 # - 3.1.0        : Add LDAP entry for 'rudder_syslog_protocol' property, and set 'tcp' as value, different from default value (udp) on a fresh 3.1
+# - 3.2.0        : Add the properties to configuration authentication provider and master admin account
 #####################################################################################
 
 # Some paths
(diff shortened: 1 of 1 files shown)


rudder-webapp/SOURCES/rudder-upgrade
```diff
@@ -397,6 +398,59 @@ rudder.cfengine.reload.server.command=/opt/rudder/bin/rudder-reload-cf-serverd"
...
+
+
+  # - 3.2.0 : Add properties to configure authentication plugins and master admin
+  # note: we can't use rudder.auth.admin.login (or .password) because it's commented
+  # out by default, so it would defeat the regex. 
+  # We also need to comment out rudder.auth.ldap.enable and use its value to decide 
+  # if we should use "ldap" of "file" for rudder.auth.type 
+
+  STEP="Add properties to configure authentication plugins and master admin"
+  if grep -iEq "rudder.auth.ldap.enable=\s*false" /opt/rudder/etc/rudder-web.properties; then
```
peckpeck: There must be a ^ at the beginning of the regex or you will match comments too.
And shouldn't there be a \s* before = too ?
Was the default value true ? If not, you should reverse the test to make the default 'file' and not ldap.

fanf: Ah, the ^ disappear somehow along the path, thanks for catching it. 
The \s* before "=" can't do any arm, I will add it to.

fanf:
PR updated

rudder-webapp/SOURCES/rudder-upgrade
```diff
@@ -397,6 +398,59 @@ rudder.cfengine.reload.server.command=/opt/rudder/bin/rudder-reload-cf-serverd"
...
+
+
+  # - 3.2.0 : Add properties to configure authentication plugins and master admin
+  # note: we can't use rudder.auth.admin.login (or .password) because it's commented
+  # out by default, so it would defeat the regex. 
+  # We also need to comment out rudder.auth.ldap.enable and use its value to decide 
+  # if we should use "ldap" of "file" for rudder.auth.type 
+
+  STEP="Add properties to configure authentication plugins and master admin"
+  if grep -iEq "^rudder.auth.ldap.enable\s*=\s*false" /opt/rudder/etc/rudder-web.properties; then
```
peckpeck: What about the default value ?
Was this line mandatory or was the default ldap ?

fanf: the line was mandatory, defaulting to rudder.auth.ldap.enable=false (so to the new "file" authentication provider).

peckpeck:
OK, so it was not possible for the line to be absent.

PR #772, opened 2015-10-30
Fixes #7335: check-rudder-agent silently fails if namespaces are not supported

peckpeck:
https://www.rudder-project.org/redmine/issues/7335


jooooooon:
I just tested this on the problematic machine and confirm it works OK.

PR #776, opened 2015-11-02
Fixes #7343: Create a migration script for relay-promises-only and cf…

ncharles:
…enigne-mission-portal parameter in the config file


rudder-webapp/SOURCES/rudder-upgrade
```diff
@@ -451,6 +452,31 @@ rudder.jdbc.maxPoolSize=25"
 rudder.auth.provider=${AUTH_PROVIDER}
   
 "
+
+  # - 3.2.0 : Add properties to define the role of servers
+  STEP="Add properties to define the new roles of servers"
+
+  check_and_add_config_property rudder.server-roles.configForRun "
```
fanf: I don't see "rudder.server-roles.configForRun", shouldn't it be "rudder.server-roles.relay-promises-only" ?

ncharles:
PR updated

peckpeck:
This is bash correct, but I think it's @fanf who should review it.

PR #783, opened 2015-11-04
Fixes #7338: All reports are missing (totally orange) for a node due to multiple cf-execd processes

peckpeck:
https://www.rudder-project.org/redmine/issues/7338


ncharles:
there is no way on earth I can review that, my bash fu is not so strong (but it does work on ubuntu)

PR #786, opened 2015-11-09
Fixes #7386: Impossible to create Technique with technique editor on Centos 6

VinceMacBuche:
https://www.rudder-project.org/redmine/issues/7386


Kegeruneku:
Okay !

PR #787, opened 2015-11-09
Fixes #7387: Do not grep rudder-jetty.conf during rudder-jetty %pre s…

Kegeruneku:
…criptlet if it does not exist

Ticket: http://www.rudder-project.org/redmine/issues/7387


peckpeck:
Why only 3.1 ?

Kegeruneku:
Because it's a 3.1 specific addition :) (it was not present before)

Kegeruneku:
(Comes from http://www.rudder-project.org/redmine/issues/6773 to be exact)

PR #789, opened 2015-11-10
Fixes #7393: Selinux policy prevent running cf-agent when using technique editor

VinceMacBuche:
https://www.rudder-project.org/redmine/issues/7393


Kegeruneku:
Looks good.

PR #790, opened 2015-11-12
Fixes #7395: After a fresh install on Centos, the Rudder server web interface does not work

VinceMacBuche:
https://www.rudder-project.org/redmine/issues/7395

### DIFF
--- a/rudder-reports/SPECS/rudder-reports.spec
+++ b/rudder-reports/SPECS/rudder-reports.spec
@@ -111,11 +111,16 @@ then
 %endif
   /sbin/service postgresql start
 fi
+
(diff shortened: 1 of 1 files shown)


rudder-reports/SPECS/rudder-reports.spec
```diff
@@ -111,11 +111,16 @@ then
 %endif
   /sbin/service postgresql start
 fi
+
+PG_HBA_FILE=$(su - postgres -c "psql -t -P format=unaligned -c 'show hba_file';")
```
ncharles: :+1:

VinceMacBuche: Thanks my dear president

PR #792, opened 2015-11-13
Fixes #7404: Merge error from #7395 - extra \"if\" in .spec pre scriplet

jooooooon:
https://www.rudder-project.org/redmine/issues/7404


rudder-reports/SPECS/rudder-reports.spec
```diff
@@ -138,9 +138,8 @@ if [ -f ${PG_HBA_FILE} ]; then
   fi
 fi
 
-  # Apply changes in PostgreSQL
-  service ${POSTGRESQL_SERVICE_NAME} reload
-fi
+# Apply changes in PostgreSQL
+service ${POSTGRESQL_SERVICE_NAME} reload
```
VinceMacBuche: Before the reload was done in the if (only if pg_hba was updated), here you will reload postgresql everytime (even if there is nothing to reload) 

Is that what you want to do ?

jooooooon:
PR updated

VinceMacBuche:
Great \o/ Merging !

PR #798, opened 2015-11-23
Fixes #7443: Call rudderify from rudder upgrade

peckpeck:
https://www.rudder-project.org/redmine/issues/7443


rudder-webapp/SOURCES/rudder-upgrade
```diff
@@ -336,6 +337,12 @@ upgrade_ncf() {
   fi
 }
 
+# Call rudderify on all local ncf techniques to make sure the promises will be properly generated
+rudderify_techniques() {
+  STEP="Call rudderify on all local ncf techniques (#7443)"
+  /usr/share/ncf/tools/ncf_rudder.py rudderify_technique /var/rudder/configuration-repository/techniques/ncf_techniques
```
VinceMacBuche: Fonction called should be rudderify_techniques !

Rudderify_technique only rudderify the technique passes as parameter!!

peckpeck:
PR updated

VinceMacBuche:
It should also reload techniques ! Two ways:

Call reload techniques API
Put force technique reload flag file so techniques will be reloded at jetty restart

ncharles:
Indeed, if the rudderify modify does something, it should call trigger_technique_library_reload

peckpeck:
PR updated

PR #804, opened 2015-11-23
Fixes #7449: rudder-upgrade recommends to use deprecated rudder-server-root init script

jooooooon:
https://www.rudder-project.org/redmine/issues/7449


peckpeck:
Merging

PR #805, opened 2015-11-24
Fixes #7453: Technique library is no longer initialised

jooooooon:
https://www.rudder-project.org/redmine/issues/7453


peckpeck:
OK Merging

PR #806, opened 2015-11-24
Fixes #7451: After an upgrade to 3.2, the rudderified techniques are not commited, and not reread

jooooooon:
https://www.rudder-project.org/redmine/issues/7451


jooooooon:
PR updated
7451

peckpeck:
OK Merging

PR #808, opened 2015-11-24
Fixes #7465: Still some isses with selinux and ncf-api-venv

VinceMacBuche:
https://www.rudder-project.org/redmine/issues/7465


VinceMacBuche:
PR updated

peckpeck:
Good, merging

PR #812, opened 2015-11-26
Fixes #7167: Put ncf/ncf_hash_file in .gitignore

jooooooon:
https://www.rudder-project.org/redmine/issues/7167

### DIFF
--- a/rudder-webapp/SOURCES/Makefile
+++ b/rudder-webapp/SOURCES/Makefile
@@ -24,7 +24,7 @@ WGET := $(if $(PROXY), http_proxy=$(PROXY) ftp_proxy=$(PROXY)) /usr/bin/wget -q
 RUDDER_VERSION_TO_PACKAGE = <put Rudder version or version-snapshot here>
 RUDDER_MAJOR_VERSION := $(shell echo ${RUDDER_VERSION_TO_PACKAGE} | cut -d'.' -f 1-2)
 
-localdepends: ./rudder-sources ./maven/bin/mvn ./rudder-users.xml ./rudder-doc ./rudder-doc/html/rudder-doc.pdf
+localdepends: ./rudder-sources ./maven/bin/mvn ./rudder-users.xml ./rudder-doc ./rudder-doc/html/rudder-doc.pdf ./.gitignore
 
 /usr/bin/wget:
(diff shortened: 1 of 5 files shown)


rudder-webapp/SOURCES/Makefile
```diff
@@ -57,11 +57,15 @@ localdepends: ./rudder-sources ./maven/bin/mvn ./rudder-users.xml ./rudder-doc .
 ./rudder-users.xml: ./rudder-sources
 	cp ./rudder-sources/rudder/rudder-web/src/main/resources/demo-rudder-users.xml ./rudder-users.xml
 
+./.gitignore: ./configuration-repository-gitignore
+	cp ./configuration-repository-gitignore ./.gitignore
```
peckpeck: Why do you copy the file here instead of just calling it with its original name in the package definition ?

rudder-webapp/SPECS/rudder-webapp.spec
```diff
@@ -103,6 +103,7 @@ Source19: rudder-reload-cf-serverd
 Source20: rudder-webapp.te
 Source21: rudder-webapp.fc
 Source22: rudder-keys
+Source23: .gitignore
```
peckpeck: Here

rudder-webapp/SOURCES/Makefile
jooooooon: I wanted to avoid committing a source file with a name beginning with a ```.```. I can see so many cases where this could cause trouble, because shells usually exclude these files from default operations (ls, cp, mv, ...). Also, I'm concerned how git would behave if it finds a ```.gitignore``` file, which is a special name for git, in a directory under git version control.

peckpeck: You can still call it configuration-repository-gitignore since there is a rename during the packaging phase (like for rudder-agent.init)

jooooooon: That is hard to acheive in debian packaging. There is a special dh_* helper that does the rename for you for init scripts (and similar things like crons), but dh_install doesn't support renaming by itself :(

peckpeck:
OK, merging

PR #816, opened 2015-11-27
Fixes #6639: Add shared-files to .gitignore

jooooooon:
https://www.rudder-project.org/redmine/issues/6639


VinceMacBuche:
I think it is useful to allow by default usage of git the shared file (stated it in redmine)

@fanf @ncharles what do you think of this ? ( @Kegeruneku and @peckpeck already gave their approval in redmine, but I'm adding them here too )

jooooooon:
As discussed in the ticket, this is a bad idea(tm).

PR #819, opened 2015-11-30
Fixes #7515: rudder-agent and rudder-slapd init scripts may fail to stop services if killall is missing or misbehaving

jooooooon:
https://www.rudder-project.org/redmine/issues/7515


peckpeck:
OK merging

PR #821, opened 2015-11-30
Fixes #7510: rudder-upgrade fails when *re*installing Rudder

jooooooon:
https://www.rudder-project.org/redmine/issues/7510


jooooooon:
Best viewed without white space change: https://github.com/Normation/rudder-packages/pull/821/files?w=1

ncharles:
\o/

PR #822, opened 2015-12-03
Fixes #4922: Rudder OpenLDAP Authentication sans ipv4 localhost

jooooooon:
https://www.rudder-project.org/redmine/issues/4922


peckpeck:
OK, merging

PR #826, opened 2015-12-07
Fixes #7420: cf-execd is running twice after a reboot or after running rudder-agent twice

peckpeck:
https://www.rudder-project.org/redmine/issues/7420

### DIFF
--- a/rudder-agent/SOURCES/rudder-agent.init
+++ b/rudder-agent/SOURCES/rudder-agent.init
@@ -191,6 +191,17 @@ start_daemons() {
 
 	# It's time to start the daemons
 	for daemon in ${DAEMONS}; do
+    # Test if daemon is already running
+    if [ -s ${CFENGINE_COMMUNITY_PID_FILE[$daemon]} ]
+    then
+      PID=`cat ${CFENGINE_COMMUNITY_PID_FILE[$daemon]}`
+      if [ -e /proc/$PID ]
+      then
+        message "info" "[INFO] CFEngine Community ${CFENGINE_COMMUNITY_NAME[$daemon]} Is already running"
+        continue
+      fi
+    fi
+
 		# Start message
 		message "info" "[INFO] Launching CFEngine Community ${CFENGINE_COMMUNITY_NAME[$daemon]}..."
 


rudder-agent/SOURCES/rudder-agent.init
```diff
@@ -191,6 +191,17 @@ start_daemons() {
 
 	# It's time to start the daemons
 	for daemon in ${DAEMONS}; do
+    # Test if daemon is already running
+    if [ -s ${CFENGINE_COMMUNITY_PID_FILE[$daemon]} ]
+    then
+      PID=`cat ${CFENGINE_COMMUNITY_PID_FILE[$daemon]}`
+      if ps -p "${PID}" > /dev/null
```
ncharles: why not using PS_COMMAND ?

rudder-agent/SOURCES/rudder-agent.init
```diff
@@ -191,6 +191,17 @@ start_daemons() {
...
+      then
+        message "info" "[INFO] CFEngine Community ${CFENGINE_COMMUNITY_NAME[$daemon]} Is already running"
+        #continue
```
jooooooon: Why is this line commented? We should uncomment it if it's needed or remove it if not.

peckpeck: Sorry, I forgot to uncomment it

rudder-agent/SOURCES/rudder-agent.init
peckpeck: PS_COMMAND is for process that have gone wild and have no pid file, it is used to detect and kill them.
Regular process run by the service should have proper pid file, the one we check here.

peckpeck:
PR updated

rudder-agent/SOURCES/rudder-agent.init
```diff
@@ -225,7 +236,7 @@ start_daemons() {
 
 		# Is daemon launched?
 		PID=`cat ${CFENGINE_COMMUNITY_PID_FILE[$daemon]}`
-		if [ ! -e /proc/$PID ]
+		if ! ps -p "${PID}" > /dev/null
```
jooooooon: Why are you changing this here? Is /proc/$PID not available on all systems the reason? Or is it related to this bug?

peckpeck:
PR updated

peckpeck:
PR updated

PR #827, opened 2015-12-07
Fixes #7423: vzps options used in proxmox are not ok

jooooooon:
https://www.rudder-project.org/redmine/issues/7423


peckpeck:
OK, merging

PR #832, opened 2015-12-08
Fixes #7031: Inventory <FQDN> content differs from hostname --fqdn and may lead to unauthorised nodes

peckpeck:
https://www.rudder-project.org/redmine/issues/7031


peckpeck:
PR updated

PR #835, opened 2015-12-09
Fixes #7581: Remove unapplying patches

jooooooon:
https://www.rudder-project.org/redmine/issues/7581


rudder-agent/SOURCES/patches/cfengine/0018-do-not-remove-cflags-when-compiling.patch
```diff
@@ -1,10 +0,0 @@
---- a/configure.ac	2015-10-12 11:48:53.451469571 +0200
-+++ b/configure.ac	2015-10-12 11:50:13.759467288 +0200
-@@ -54,7 +54,6 @@
- dnl It is restored later.
- dnl
- ENV_CFLAGS="$CFLAGS"
--CFLAGS=
```
jooooooon: This line is no longer in CFEngine 3.6.3's configure.ac and I don't see any other lines in there that reset CFLAGS, so I assume this patch is no longer necessary.

rudder-agent/SOURCES/patches/cfengine/07-define-classes-when-package-is-up-to-date.patch
```diff
@@ -1,13 +0,0 @@
-diff -ur a/cf-agent/verify_packages.c b/cf-agent/verify_packages.c
```
jooooooon: This is a duplicate patch with 0017-define-classes-when-package-is-up-to-date.patch. They do the same thing, therefore the second doesn't apply, causing errors.

jooooooon:
Don't merge yet

jooooooon:
In 3.2 branch the *other* patch for define-classes-when-package-is-up-to-date was removed, so we should remove the same one to avoid the merge ending up with neither

jooooooon:
PR updated

rudder-agent/SOURCES/patches/cfengine/0017-define-classes-when-package-is-up-to-date.patch
```diff
@@ -1,13 +0,0 @@
-diff -ur a/cf-agent/verify_packages.c b/cf-agent/verify_packages.c
```
jooooooon: This patch is a duplicate of patch 07-define-classes-when-package-is-up-to-date.patch and has already been removed in the 3.2 branch.

jooooooon:
OK, good to go again.

peckpeck:
OK merging

PR #838, opened 2015-12-17
Fixes #7630: Backport cfengine networking fixes for Rudder 3.0

amousset:
http://www.rudder-project.org/redmine/issues/7630


ncharles:
I am so happy !

PR #839, opened 2015-12-17
Fixes #7632: Backport the patch to shutdown the socket in case of tim…

amousset:
…eout for 3.0

http://www.rudder-project.org/redmine/issues/7632


amousset:
Do not merge for now, this patch does not work.

amousset:
PR updated with a fixed patch.

PR #844, opened 2015-12-22
Fixes #7677: Error in rudder-agent init script on ubuntu - missing fix

amousset:
https://www.rudder-project.org/redmine/issues/7677


peckpeck:
OK, merging

PR #845, opened 2015-12-28
Fixes #7679: All the files in /opt/rudder/share/tools should not be executable

jooooooon:
https://www.rudder-project.org/redmine/issues/7679


peckpeck:
OK, merging

peckpeck:
Not merging, maybe we should put it in 3.2 only

jooooooon:
Replaced by #846

PR #846, opened 2015-12-28
Fixes #7679: All the files in /opt/rudder/share/tools should not be executable

jooooooon:
https://www.rudder-project.org/redmine/issues/7679


peckpeck:
So this entire files does nothing excepting checking its arguments.
Shouldn't it be simply removed ?

jooooooon:
Ah, yes, indeed it should. Well spotted, thanks.

jooooooon:
PR updated

peckpeck:
OK, merging

PR #848, opened 2015-12-31
Fixes #7710: Broken builds due to undefined variable

amousset:
https://www.rudder-project.org/redmine/issues/7710


peckpeck:
OK, merging

PR #849, opened 2016-01-02
Fixes #7667: Disable ldap logging on non existing indexes

jooooooon:
https://www.rudder-project.org/redmine/issues/7667


peckpeck:
Shouldn't this be done in 2.11 ?

jooooooon:
This is not the right approach: it would remove *all* logs from slapd.

PR #850, opened 2016-01-04
Fixes #7718: Put a makefile inside each directory of rudder-packages

peckpeck:
https://www.rudder-project.org/redmine/issues/7718

### DIFF
--- a/packages.makefile
+++ b/packages.makefile
@@ -1,0 +1,98 @@
+#####################################################################################
+# Copyright 2011 Normation SAS
+#####################################################################################
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, Version 3.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+#####################################################################################
+
+include $(TOOLS_DIR)/build-package.makefile
+
+ARCHI := $(shell uname -m)
+OS := $(shell . ../detect_os.sh && echo $${OS})
+OSSP := $(shell . ../detect_os.sh && echo $${OSSP})
+OSVERSION := $(shell . ../detect_os.sh && echo $${OSVERSION})
+
+BUILDREQUIRESSLES := $(shell grep -s "SLES${OSVERSION}:" SOURCES/.dependencies | cut -d ':' -f2)
+BUILDREQUIRESSLESSP := $(shell grep -s "SLES${OSVERSION}SP${OSSP}:" SOURCES/.dependencies | cut -d ':' -f2)
+BUILDREQUIRESRHEL := $(shell grep -s "RHEL${OSVERSION}" SOURCES/.dependencies | cut -d ':' -f2)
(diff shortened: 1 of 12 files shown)


peckpeck:
We keep the name package.makefile because it reminds us its origin and not Makefile because it is not meant to be run from the root path.
The file is a direct copy from packaging branch 2.11

packages.makefile
```diff
@@ -0,0 +1,98 @@
...
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+#####################################################################################
+
+include $(TOOLS_DIR)/build-package.makefile
+
+ARCHI := $(shell uname -m)
+OS := $(shell . SOURCES/rudder-sources/rudder-packages/detect_os.sh && echo $${OS})
+OSSP := $(shell . SOURCES/rudder-sources/rudder-packages/detect_os.sh && echo $${OSSP})
+OSVERSION := $(shell . SOURCES/rudder-sources/rudder-packages/detect_os.sh && echo $${OSVERSION})
```
jooooooon: Shouldn't we adapt this path to use the local copy of detect_os.sh ?

packages.makefile
```diff
@@ -0,0 +1,98 @@
...
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+#####################################################################################
+
+include $(TOOLS_DIR)/build-package.makefile
```
jooooooon: Why not include this file too?

jooooooon: By "include" I mean "copy it to rudder-packages"

rudder-policy-templates/Makefile
```diff
@@ -0,0 +1 @@
+include ../packages.makefile
```
jooooooon: Wow. This package should NOT still be around - it was deprecated in Rudder 2.4!

jooooooon: (not to worry for this PR, of course, but we should remove it separately)

packages.makefile
peckpeck: I don't know where TOOL directory is nor how the file is copied there.
So I leaved everything as it was before.

packages.makefile
peckpeck: This path will soon be a local copy, cf #851

rudder-policy-templates/Makefile
jooooooon: I've created https://www.rudder-project.org/redmine/issues/7723 for this

packages.makefile
jooooooon: As agreed, we'll make sure to change the path here to ../detect_os.sh, once we have that file copied to all our builders.

packages.makefile
peckpeck: Some of its content is Normation specific.
I think I will merge it back progressively directly into this file

peckpeck:
PR updated

peckpeck:
The patch to copy detect_os .sh and friends has been merged so this can be merged too

rudder-policy-templates/Makefile
jooooooon: Since we've removed this package, you should remove this Makefile from your PR.

peckpeck:
PR updated

jooooooon:
I have finally finished reviewing this. My apologies for the very long delay on this one!

I think this is fine, however I have noticed that there are some changes to this Makefile in branches 3.0 (change handling and versions of JDK) and 3.2 (addition of Solaris support). We will need extra Pull Requests to add these changes back in, for those two branches.

Merging this as is for now.

PR #851, opened 2016-01-04
Fixes #7721: Add detect_os to rudder-agent's package source

peckpeck:
https://www.rudder-project.org/redmine/issues/7721


jooooooon:
As discussed this doesn't make sense

PR #855, opened 2016-01-08
Fixes #7620: rudder-agent should trigger a new inventory after upgrade

amousset:
https://www.rudder-project.org/redmine/issues/7620

### DIFF
--- a/rudder-agent/SPECS/rudder-agent.spec
+++ b/rudder-agent/SPECS/rudder-agent.spec
@@ -646,6 +646,14 @@ if [ -f ${TMP_CRON} ]; then
 fi
 %endif
 
+# Try to send an inventory after upgrade to see the new agent version on the server
+if [ ${CFRUDDER_FIRST_INSTALL} -ne 1 ]
+then
+  echo "INFO: Trigger an inventory..."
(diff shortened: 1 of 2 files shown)


rudder-agent/SPECS/rudder-agent.spec
```diff
@@ -646,6 +646,14 @@ if [ -f ${TMP_CRON} ]; then
 fi
 %endif
 
+# Try to send an inventory after upgrade to see the new agent version on the server
+if [ ${CFRUDDER_FIRST_INSTALL} -ne 1 ]
+then
+  echo "INFO: Trigger an inventory..."
+  /var/rudder/cfengine-community/bin/cf-agent -KD force_inventory >> %{rudderlogdir}/install/rudder-agent.log 2>&1
```
jooooooon: I assume this will be changed to "rudder agent inventory" in versions where that command exists?

amousset: Yes, this version is only for 2.11, it will be `/opt/rudder/bin/rudder agent inventory` for other versions.

jooooooon: Great.

PR #858, opened 2016-01-14
Fixes #7770: If the inventory fails during install, the whoe install fails

peckpeck:
https://www.rudder-project.org/redmine/issues/7770


jooooooon:
:+1:

jooooooon:
Ah, but we should do the same change in the .spec file. I understand that as of RHEL >= 7, rpm also stops on non-zero return codes.

peckpeck:
PR updated

PR #863, opened 2016-01-18
Fixes #7797: Warn when the servers answers an empty string to a (S)OPENDIR

amousset:
https://www.rudder-project.org/redmine/issues/7797


amousset:
PR updated

jooooooon:
Looks good. However, this will get merged up into the 3.2 branch, where the patch won't apply.

Once I've merged this PR, please run "rudder-dev merge all" on your workstation and then prepare another PR to remove the patch from 3.2.

PR #867, opened 2016-01-19
Fixes #7803: Test for non-applying patches on rudder-jetty and rudder-agent-thin is broken

jooooooon:
https://www.rudder-project.org/redmine/issues/7803


Kegeruneku:
OK !

PR #868, opened 2016-01-19
Fixes #4473: Jetty cache files in /tmp directory are removed by the OS tmpwatch cron job, preventing access to the application

jooooooon:
https://www.rudder-project.org/redmine/issues/4473


Kegeruneku:
Accepted !

PR #877, opened 2016-01-27
Fixes #7846: Default vhost is not disabled on rudder-relay 

VinceMacBuche:
https://www.rudder-project.org/redmine/issues/7846

### DIFF
--- a/rudder-server-relay/debian/postinst
+++ b/rudder-server-relay/debian/postinst
@@ -21,7 +21,7 @@ set -e
 case "$1" in
     configure)
 
-      SITES_TO_DISABLE="default default-ssl"
+      SITES_TO_DISABLE="default 000-default default-ssl"
       SITES_TO_ENABLE="rudder-relay-vhost"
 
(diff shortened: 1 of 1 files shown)


rudder-server-relay/debian/postinst
```diff
@@ -21,7 +21,7 @@ set -e
 case "$1" in
     configure)
 
-      SITES_TO_DISABLE="default default-ssl"
+      SITES_TO_DISABLE="default 000-default default-ssl"
```
jooooooon: a2dissite doesn't normally work like this...

jooooooon: Ahhhh, I think what is happening here is that the *name* of the default site has been changed between Apache 2.2 and 2.4:

Apache 2.2:
sites-available/default.conf

Apache 2.4:
sites-available/000-default.conf

Therefore, one out of two of these disable commands will fail, because only one exists. However, that is fine, because the a2dissite command below is followed by ```|| true```

PR #878, opened 2016-01-27
Fixes #7849: rudder-upgrade does not make some check on remote sql in case of distributed setup

ncharles:
https://www.rudder-project.org/redmine/issues/7849


peckpeck:
OK, merging

PR #879, opened 2016-02-04
Fixes #4499: Rudder init script kill all agent on Open VZ (or similar system)

amousset:
https://www.rudder-project.org/redmine/issues/4499


peckpeck:
OK, merging

PR #882, opened 2016-02-11
Fixes #7902: Add support to cfengine bootstrap in server

peckpeck:
https://www.rudder-project.org/redmine/issues/7902


jooooooon:
:+1: 

This is great, but will only work for upgrades (rudder-upgrade is, as it's name implies, only run on upgrades). You should add something similar in post-install to do the rsync for initial installations. In fact, that would make more sense, in the rudder-techniques postinstall, and then that's the only place you need it.

jooooooon:
Hmmm, why is this targeted to 3.0? That doesn't make any sense to me, either we're doing something safe and with pretty much no impact and it can go into 2.11, or it's a risk and it should be on master.

peckpeck:
The problem with post-install is that we don't know if the agent has been installed yet or not.

It is targeted to old versions because it needs to be present on servers the day we create an agent without initial promises.
It is not targeted to 2.11 because nch think it should not impact this old release.

jooooooon:
> The problem with post-install is that we don't know if the agent has been installed yet or not.

Why is that important? Either way, we can't introduce a change that only works on upgrades and not on new installs.

> It is targeted to old versions because it needs to be present on servers the day we create an agent without initial promises.

I understand the requirement to provide forward-compatibility, but this is an exception to our release maintenance policy, and we should discuss it and address it as such. Pull Requests are really not a good place for such discussions.

> It is not targeted to 2.11 because nch think it should not impact this old release.

This makes no sense: 2.11, 3.0 and 3.1 are in exactly the same state of maintenance - released, stable versions, that should see the least possible changes in order to preserve and grow that stability.

Either we are concerned about this change breaking something, in which case it should go to master as per our standard policy, or we are willing to make an exception (see above) because we evaluate the risk to be lower than the benefit, but in that case there is no reason to exclude one version out of all the ones we maintain.

peckpeck:
it important because if the agent is not installed, the initial promises to copy are not there.
Moreover rudder-upgrade is always run, whether at first install or during upgrade.

jooooooon:
> Moreover rudder-upgrade is always run, whether at first install or during upgrade.

Wow, you're absolutely right about that. Sorry, I was *convinced* that was not the case, due to a support issue recently, but that was for another reason.

> it important because if the agent is not installed, the initial promises to copy are not there.

Right, indeed. This problem is the same though for rudder-upgrade, since it is run from post-install. I think that's good enough for a packaging implementation, though, but maybe we should double this with a CFEngine copy if it's important?

rudder-webapp/SOURCES/rudder-upgrade
```diff
@@ -359,6 +360,13 @@ upgrade_ncf() {
 
 }
 
+# Upgrade masterfiles
+upgrade_masterfiles() {
+  # - 3.0.13  Now we should have a masterfiles initialized from initial promises
+  STEP="Upgrade masterfiles from initial promises"
+  rsync --delete -rptgoq "${RUDDER_SHARE}/initial-promises/" "${RUDDER_VAR}/cfengine-community/masterfiles/"
```
jooooooon: I think we should add a ```mkdir -p "${RUDDER_VAR}/cfengine-community/masterfiles/"``` before the rsync, to make absolutely sure this can't cause an error on upgrade - especially if we include it in existing released branches.

It is *possible* that this can be executed before the agent is installed, therefore before that directory exists.

jooooooon: Actually, re-reading the comments (in particular: "if the agent is not installed, the initial-promises aren't there"), we need a different protection on this: test that ```"${RUDDER_SHARE}/initial-promises/"``` exists.

peckpeck:
PR updated

peckpeck:
PR updated

rudder-webapp/SOURCES/rudder-upgrade
```diff
@@ -369,6 +370,17 @@ upgrade_ncf() {
 
 }
 
+# Upgrade masterfiles
+upgrade_masterfiles() {
+  # - 3.0.13  Now we should have a masterfiles initialized from initial promises
```
jooooooon: Version needs updating

jooooooon:
Please retarget to 3.1 as discussed

peckpeck:
PR replaced by https://github.com/Normation/rudder-packages/pull/907

PR #883, opened 2016-02-12
Fixes #7901: Backup and restore policy_server.dat like we do with uuid.hive file

jooooooon:
https://www.rudder-project.org/redmine/issues/7901


peckpeck:
Good, merging

PR #891, opened 2016-02-29
Fixes #7777: Don't provide the same capability as the unversioned rpm

matya:
Fixes #7777: Don't provide the same capability as the unversioned rpm

### DIFF
--- a/rudder-agent/SPECS/rudder-agent.spec
+++ b/rudder-agent/SPECS/rudder-agent.spec
@@ -89,7 +89,7 @@ BuildRoot: %{_tmppath}/%{name}-%{version}-%{release}-root-%(%{__id_u} -n)
 # Generic requirements
 BuildRequires: gcc bison flex pcre-devel autoconf automake libtool
 Requires: pcre
-Provides: rudder-agent
+Provides: Rudder::Agent
 Conflicts: rudder-agent-thin
 
 # Specific requirements
--- a/rudder-server-root/SPECS/rudder-server-root.spec
+++ b/rudder-server-root/SPECS/rudder-server-root.spec
@@ -52,7 +52,7 @@ Source8: rudder-server-root
 BuildRoot: %{_tmppath}/%{name}-%{version}-%{release}-root-%(%{__id_u} -n)
 BuildArch: noarch
 
-Requires: rudder-jetty, rudder-webapp, rudder-inventory-endpoint, rudder-inventory-ldap, rudder-reports, rudder-agent, curl
+Requires: rudder-jetty, rudder-webapp, rudder-inventory-endpoint, rudder-inventory-ldap, rudder-reports, Rudder::Agent, curl
 
 %description
 Rudder is an open source configuration management and audit solution.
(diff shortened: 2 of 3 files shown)


rudder-server-root/SPECS/rudder-server-root.spec
```diff
@@ -52,7 +52,7 @@ Source8: rudder-server-root
 BuildRoot: %{_tmppath}/%{name}-%{version}-%{release}-root-%(%{__id_u} -n)
 BuildArch: noarch
 
-Requires: rudder-jetty, rudder-webapp, rudder-inventory-endpoint, rudder-inventory-ldap, rudder-reports, rudder-agent, curl
+Requires: rudder-jetty, rudder-webapp, rudder-inventory-endpoint, rudder-inventory-ldap, rudder-reports, Rudder::Agent, curl
```
peckpeck: Will rpm automatically install rudder-agent if it's not already there with this dependency ?
For example, on debian you should add an "or" with a real package here.

matya: It will resolve it as a dependency, and either install the the regular or their thin client. Not sure if you can have a weight which to choose.

rudder-agent/SPECS/rudder-agent.spec
```diff
@@ -89,7 +89,7 @@ BuildRoot: %{_tmppath}/%{name}-%{version}-%{release}-root-%(%{__id_u} -n)
 # Generic requirements
 BuildRequires: gcc bison flex pcre-devel autoconf automake libtool
 Requires: pcre
-Provides: rudder-agent
+Provides: Rudder::Agent
```
jooooooon: Shouldn't this "Provides" line also include %{version}? My understanding is that we will hit the same problem again if nto.

rudder-server-root/SPECS/rudder-server-root.spec
jooooooon: I'm a bit concerned by this, because if it ends up asking the user to make a choice, we're losing out on ease-of-install. On Debian, this would be the case, but I'm not sure on SLES/RHEL.

matya: Well, something has to be done, this war zypper willing not update rudder-agent on sles... I have to repackage basically all rudder-agent packages by http://rpmrebuild.sourceforge.net

rudder-agent/SPECS/rudder-agent.spec
matya: That "provides" is done implicitly by rpmbuild.

rudder-server-root/SPECS/rudder-server-root.spec
peckpeck: The or operator is supported for dependencies as show here : 
https://docs.fedoraproject.org/en-US/Fedora_Draft_Documentation/0.1/html/RPM_Guide/ch-specfile-syntax.html#id729642

So you could use "Rudder::Agent or rudder-agent"

jooooooon:
@matya May I ask why you closed this PR? I know this hasn't moved forwards in a while, but this is because it requires significant testing (we really don't want to make things worse!).

I think this approach is probably the right one, we just need to test it. @amousset will be doing that.

Please let us know if you closed this for some other reason?

matya:
Sorry for closing it abruptly without comments, but I lack the time to continue working with this issue, and this pr is untested, so not ready for merge. 

Please feel free to just use the patch suggested for testing, but please do it on a separate repo, so possible fixes can be implemented without me having to write them based on your test results.

amousset:
Commit modified

PR #903, opened 2016-03-29
Fixes #8117: Error in relay post install script on debian

amousset:
https://www.rudder-project.org/redmine/issues/8117


amousset:
Same as #892 but for the relay.

peckpeck:
Good, merging

PR #904, opened 2016-03-31
Fixes #8054: node-to-relay fails to reload policy when proxy is set

peckpeck:
https://www.rudder-project.org/redmine/issues/8054


amousset:
Shouldn't we add this to all HTTP calls to local API (rudder server reload-*, the /status call in rudder agent, etc.) ?

peckpeck:
yes, but most of them should already be done.
Can you find one ?

amousset:
I see:
- https://github.com/Normation/rudder-techniques/blob/master/techniques/system/server-roles/1.0/technique-reload.st#L60
- https://github.com/Normation/rudder-techniques/blob/master/initial-promises/node-server/server-roles/1.0/alive-check.cf#L57
- https://github.com/Normation/rudder-agent/pull/54/files#diff-ab7a4a234a368e74ff367371269d02dcR7

peckpeck:
PR updated

amousset:
Ok, merging.

PR #909, opened 2016-04-29
Fixes #7667: Disable ldap logging on non existing indexes

amousset:
https://www.rudder-project.org/redmine/issues/7667


jooooooon:
:+1:

PR #910, opened 2016-04-29
Fixes #8223: Rudder 3.3 doesn't build on RHEL3

peckpeck:
https://www.rudder-project.org/redmine/issues/8223

### DIFF
--- a/rudder-agent/SOURCES/Makefile
+++ b/rudder-agent/SOURCES/Makefile
@@ -27,6 +27,7 @@ FUSION_RELEASE = 2.3.6
 LMDB_RELEASE = 0.9.16
 OPENSSL_RELEASE = 1.0.1p
 PERL_VERSION = 5.22.0
+PCRE_RELEASE = 8.38
 
 # Defauth PATHs
 RUDDER_DIR = /opt/rudder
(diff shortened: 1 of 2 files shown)


rudder-agent/SOURCES/Makefile
```diff
@@ -316,6 +332,11 @@ build-lmdb: lmdb-source
 	cd build-lmdb/opt/rudder && mkdir -p bin lib include man/man1
 	cd lmdb-source/libraries/liblmdb && $(MAKE) install prefix=$(RUDDER_DIR) DESTDIR=$(CURDIR)/build-lmdb
 
+pcre-lmdb: pcre-source
```
jooooooon: I think you meant to call this target build-pcre

rudder-agent/SOURCES/Makefile
```diff
@@ -316,6 +332,11 @@ build-lmdb: lmdb-source
...
+	cd pcre-source && ./configure --disable-cpp --enable-utf8 --enable-unicode-properties --prefix=$(RUDDER_DIR)
+	cd pcre-source && $(MAKE)
+	cd pcre-source && $(MAKE) install DESTDIR=$(CURDIR)/build-pcre
```
jooooooon: Why do you do the make install in the build target, not the install target below?

rudder-agent/SOURCES/Makefile
peckpeck: yes

rudder-agent/SOURCES/Makefile
peckpeck: The make install is done in a temporary location because it is needed to build cfengine
The install target then simply copy the files to their final location
This is the same process as for lmdb and openssl

peckpeck:
PR updated

rudder-agent/SOURCES/Makefile
jooooooon: Ah, I see. Since this not obvious, cuold you add a comment to explain this succinctly in the code?

jooooooon:
OK, thanks, looks good!

PR #912, opened 2016-05-02
Fixes #6468: When we install rudder-agent after having removing it, the uuid is not restored on RPM-based systems

jooooooon:
https://www.rudder-project.org/redmine/issues/6468

### DIFF
--- a/rudder-agent/SPECS/rudder-agent.spec
+++ b/rudder-agent/SPECS/rudder-agent.spec
@@ -495,20 +495,6 @@ if ! grep -q "/opt/rudder/lib" /etc/ld.so.conf; then
 fi
 %endif
 
-# Generate a UUID if we don't have one yet
-if [ ! -e /opt/rudder/etc/uuid.hive ]
-then
-  uuidgen > /opt/rudder/etc/uuid.hive
-else
-  # UUID is valid only if it has been generetaed by uuidgen or if it is set to 'root' for policy server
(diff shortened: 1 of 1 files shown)


rudder-agent/SPECS/rudder-agent.spec
```diff
@@ -495,20 +495,6 @@ if ! grep -q "/opt/rudder/lib" /etc/ld.so.conf; then
...
-else
-  # UUID is valid only if it has been generetaed by uuidgen or if it is set to 'root' for policy server
-  CHECK_UUID=`cat /opt/rudder/etc/uuid.hive | grep -E "^[a-z0-9]{8}-[a-z0-9]{4}-[a-z0-9]{4}-[a-z0-9]{4}-[a-z0-9]{12}|root" | wc -l`
-  # If the UUID is not valid, regenerate it
-  if [ ${CHECK_UUID} -ne 1 ]
-  then
-    uuidgen > /opt/rudder/etc/uuid.hive
-  fi
-fi
-
```
jooooooon: Removing this because the exact same code is done in the check-rudder-agent command, that is run at the end of postinst. See http://www.rudder-project.org/redmine/issues/6468#note-18 for rationale.

jooooooon:
Note this change is for the .spec only, because .deb postinst doesn't have this section.

peckpeck:
Beware that in master this code is in SOURCES/rudder-agent-postinst so it will be properly merged but not removed.

peckpeck:
OK, you can self merge via rudder-dev

PR #913, opened 2016-05-02
Fixes #8226: Adapt rudder agent thin to the new package makefile

peckpeck:
https://www.rudder-project.org/redmine/issues/8226


rudder-agent-thin/SOURCES/Makefile
```diff
@@ -1,79 +1,23 @@
...
+
+# when a Makefile regenerates itself, make reexecute itself on the new makefile
+Makefile: always
+	rsync -a ../../rudder-agent/ ../
 	# Convert rudder-agent to rudder-agent-thin where needed
-	$(PATCH) -d ../debian --forward -p1 < patches/rudder-agent-thin/0001-rudder-agent-to-rudder-agent-thin-debian.patch
-	$(PATCH) -d ../SPECS --forward -p1 < patches/rudder-agent-thin/0002-rudder-agent-to-rudder-agent-thin-spec.patch
-
-	# Make sure there were no rejects
-	test `$(FIND) . -name \*.rej | wc -l` = 0
```
jooooooon: Please keep this test. We have had a lot of trouble with partially/completely failed patches that we thought were applied but weren't... patch doesn't return a useful error code, so we have to use this instead

rudder-agent/SPECS/rudder-agent.spec
```diff
@@ -67,7 +76,9 @@ Source4: perl-prepare.sh
 Source5: rudder-agent.default
 Source6: rudder-agent.init
 Source7: rudder-agent.sh
+%if %{real_name} == "rudder-agent"
```
jooooooon: Why not use %{use_system_perl} for this condition? It seems more logical to me

rudder-agent/SPECS/rudder-agent.spec
```diff
@@ -78,20 +89,31 @@ Source15: vzps.py
 Source16: rudder.8.gz
 Source17: signature_check.sh
 
+%if %{real_name} == "rudder-agent"
```
jooooooon: Same question here, this is directly related to building our own perl

rudder-agent/SPECS/rudder-agent.spec
```diff
@@ -256,7 +280,7 @@ cd %{_sourcedir}
 export CFLAGS="${RPM_OPT_FLAGS}"
 export CXXFLAGS="${RPM_OPT_FLAGS}"
 
-make %{?_smp_mflags} USE_SYSTEM_OPENSSL=%{use_system_openssl} USE_SYSTEM_LMDB=%{use_system_lmdb} USE_SYSTEM_PCRE=%{use_system_pcre}
+make %{?_smp_mflags} USE_SYSTEM_OPENSSL=%{use_system_openssl} USE_SYSTEM_LMDB=%{use_system_lmdb} USE_SYSTEM_PCRE=%{use_system_pcre} USE_SYSTEM_FUSION=%{use_system_fusion} USE_SYSTEM_perl=%{use_system_perl}
```
jooooooon: This should be USE_SYSTEM_PERL (missing caps on perl)

rudder-agent/SPECS/rudder-agent.spec
```diff
@@ -275,12 +299,12 @@ cd %{_sourcedir}
 %define no_profile true
 %endif
 
-make install DESTDIR=%{buildroot} USE_SYSTEM_OPENSSL=%{use_system_openssl} USE_SYSTEM_LMDB=%{use_system_lmdb} USE_SYSTEM_PCRE=%{use_system_pcre} NO_INIT=%{no_init} NO_CRON=%{no_cron} NO_LD=%{no_ld} NO_PROFILE=%{no_profile} 
+make install DESTDIR=%{buildroot} USE_SYSTEM_OPENSSL=%{use_system_openssl} USE_SYSTEM_LMDB=%{use_system_lmdb} USE_SYSTEM_PCRE=%{use_system_pcre} NO_INIT=%{no_init} NO_CRON=%{no_cron} NO_LD=%{no_ld} NO_PROFILE=%{no_profile} USE_SYSTEM_FUSION=%{use_system_fusion} USE_SYSTEM_perl=%{use_system_perl}
```
jooooooon: This should be USE_SYSTEM_PERL (missing caps on perl)

rudder-agent-thin/SOURCES/Makefile
```diff
@@ -1,79 +1,23 @@
...
-
-	touch build-sources-stamp
+	patch -d ../debian --forward -p1 < patches/rudder-agent-thin/0001-rudder-agent-to-rudder-agent-thin-debian.patch
```
jooooooon: You removed this patch, why is it still here?

rudder-agent-thin/SOURCES/Makefile
jooooooon: Actually, this depends whether you left the patch command below in by mistake or not :)

rudder-agent-thin/SOURCES/Makefile
```diff
@@ -1,79 +1,23 @@
...
-# We need to copy rudder-agent's Makefile to make the build the most self-contained possible
-rudder-agent.make: ../../rudder-agent/SOURCES/Makefile
-	cp -f ../../rudder-agent/SOURCES/Makefile ./rudder-agent.make
-
-fetch-sources: fetch-sources-stamp
-fetch-sources-stamp:
-	# Clean up rudder-agent sources first
-	cd ../../rudder-agent/SOURCES && make localclean
-	# Make sure the debian/changelog is pristine
-	cd ../../rudder-agent/debian && git checkout changelog
```
jooooooon: Shouldn't we keep this hack?

rudder-agent/debian/control
```diff
@@ -8,7 +8,7 @@ Homepage: http://www.rudder-project.org
 
 Package: rudder-agent
 Architecture: any
-Depends: ${shlibs:Depends}, ${misc:Depends}, uuid-runtime, dmidecode, cron, net-tools, openssl
+Depends: ${shlibs:Depends}, ${misc:Depends}, uuid-runtime, dmidecode, cron, net-tools, openssl ${rudderDepends}
```
jooooooon: I like the look of this, but I don't understand where this variable comes from

rudder-agent-thin/SOURCES/Makefile
peckpeck: No, the packaging will call it directly
And since I used the Makefile restart hack, it will be called directly on the patched rudder-agent copy.

rudder-agent/debian/control
peckpeck: It's a mistake, I tried to make this file dynamically built, but it need a lots of variables to work so I abandoned the idea

peckpeck:
PR updated

jooooooon:
OK, this looks good! Merging.

PR #921, opened 2016-05-11
Fixes #8269: Adapt rudder agent thin to the new package makefile

peckpeck:
https://www.rudder-project.org/redmine/issues/8269


rudder-agent-thin/SOURCES/Makefile
```diff
@@ -16,7 +16,9 @@ Makefile: always
 	patch -d ../debian --forward -p1 < patches/rudder-agent-thin/0001-rudder-agent-to-rudder-agent-thin-debian.patch
 	test `find . -name \*.rej | wc -l` = 0
 	sed -i 's/^REAL_NAME=rudder-agent/REAL_NAME=rudder-agent-thin/' ../debian/rules
+	sed -i 's/^rudder-agent/rudder-agent-thin/' ../debian/changelog
 	sed -i 's/^%define real_name  *rudder-agent$$/%define real_name rudder-agent-thin/' ../SPECS/rudder-agent.spec
+	mv ../debian/rudder-agent.install ../debian/rudder-agent-thin.install
```
jooooooon: It would be simpler to rename this file just "debian/install". ```dh_install``` looks by default for a file named "${package}.install", but falls back to "install". Since we have two different package names, let's just use the common name.

See https://www.debian.org/doc/manuals/maint-guide/dother.en.html#install for details.

peckpeck:
Commit modified

rudder-agent-thin/SOURCES/Makefile
peckpeck: I like it

PR #922, opened 2016-05-11
Fixes #8272: ACL Posix on Git repos  - fix bug in case command is missing

jooooooon:
https://www.rudder-project.org/redmine/issues/8272


peckpeck:
OK, merging

PR #928, opened 2016-05-15
Fixes #8295: rudder init script reports wrong status if run as non-root

jooooooon:
http://www.rudder-project.org/redmine/issues/8295


jooooooon:
Some config issues with rudder-dev have caused this PR to be created from a branch pushed to the origin (Normation) repo. Sorry. Closing.

jooooooon:
Replaced by #930

PR #929, opened 2016-05-15
Fixes #8295: rudder init script reports wrong status if run as non-root

jooooooon:
https://www.rudder-project.org/redmine/issues/8295


jooooooon:
Another failed PR sorry

jooooooon:
Replaced by #930

PR #931, opened 2016-05-15
Fixes #8296: rudder init script fails with \"/etc/init.d/rudder: line 109: printf \033[1;32m---8<---\033[0;39m\n: command not found\"

jooooooon:
https://www.rudder-project.org/redmine/issues/8296


peckpeck:
Good, merging

PR #932, opened 2016-05-16
Fixes #8231: rudder-apache gets added multiple times to apache config

jooooooon:
https://www.rudder-project.org/redmine/issues/8231

### DIFF
--- a/rudder-webapp/SPECS/rudder-webapp.spec
+++ b/rudder-webapp/SPECS/rudder-webapp.spec
@@ -286,7 +286,6 @@ echo " Done"
 # Do this ONLY at first install
 if [ $1 -eq 1 ]
 then
-		echo -e '# This sources the configuration file needed by Rudder\n. /etc/sysconfig/rudder-apache' >> /etc/sysconfig/apache2
 		echo 'DAVLockDB /tmp/davlock.db' > /etc/%{apache}/conf.d/dav_mod.conf
 fi
 


rudder-webapp/SPECS/rudder-webapp.spec
```diff
@@ -286,7 +286,6 @@ echo " Done"
 # Do this ONLY at first install
 if [ $1 -eq 1 ]
 then
-		echo -e '# This sources the configuration file needed by Rudder\n. /etc/sysconfig/rudder-apache' >> /etc/sysconfig/apache2
```
jooooooon: This is already done just a few lines below, with a safeguard to avoid adding it twice

peckpeck: It is done between a  %if 0%{?sles_version} so it shoundprobably be changed too.

jooooooon: I don't understand what you're saying, sorry. The alternative version below if in a %if statement to make sure it's only on SLES, yes, because /etc/sysconfig/apache2 is SLES-specific.

What do you want me to change?

peckpeck: OK, it's just that I was confused by the fact you  are removing this line in a non SLES specific section.
So we have created a useless file on redhat during the last few versions

peckpeck:
OK, merging

rudder-webapp/SPECS/rudder-webapp.spec
jooooooon: Yes, indeed :)

PR #935, opened 2016-05-18
Fixes #8232: Apache config does not get cleaned up after removal of rudder-webapp

jooooooon:
https://www.rudder-project.org/redmine/issues/8232

### DIFF
--- a/rudder-webapp/SPECS/rudder-webapp.spec
+++ b/rudder-webapp/SPECS/rudder-webapp.spec
@@ -436,6 +436,18 @@ if [ $1 -eq 0 ]; then
     groupdel %{config_repository_group}
     echo " Done"
   fi
+
+%if 0%{?sles_version}
+  # Remove required includes in the SLES apache2 configuration
+  if [ -f /etc/sysconfig/apache2 ]; then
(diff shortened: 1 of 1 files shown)


rudder-webapp/SPECS/rudder-webapp.spec
```diff
@@ -436,6 +436,18 @@ if [ $1 -eq 0 ]; then
     groupdel %{config_repository_group}
     echo " Done"
   fi
+
+%if 0%{?sles_version}
+  # Remove required includes in the SLES apache2 configuration
+  if grep -qE "^. /etc/sysconfig/rudder-apache$" /etc/sysconfig/apache2; then
```
peckpeck: The grep is already done by the sed line since sed won't do anything if there is no change.
Shouldn't we check the existence of the file instead ? (in case the user remove apache at the same time)

jooooooon: Yes, good point and good idea!

jooooooon:
Commit modified

peckpeck:
Good, merging

PR #938, opened 2016-05-18
Fixes #7348: rudder-jetty initialisation should not support Java 6

jooooooon:
https://www.rudder-project.org/redmine/issues/7348


peckpeck:
OK, merging

PR #940, opened 2016-05-18
Fixes #7833: Upgrade to CFEngine 3.7.3

amousset:
https://www.rudder-project.org/redmine/issues/7833


amousset:
Commit modified

amousset:
We are almost even with upstream! We still have to keep the patch for the runlog file (it has only been integrated upsteam in 3.8).

Added the patches for:
* default bodies
* callstack function

I also integrated the tests for these patches, which was not always for previous Rudder releases. We should be able to run as many CFEngine tests as possible against our agent.

ncharles:
This is awesome ! 
Just a question: shouldn't we backport the callstack function in 3.2 as well ?

amousset:
PR rebased

amousset:
Commit modified

peckpeck:
The callstack triggers a significant amount of code in ncf 1.0, I don't think we should enable it in a minor release

jooooooon:
peckpeck wrote:
> The callstack triggers a significant amount of code in ncf 1.0, I don't think we should enable it in a minor release

This is targeted to master, so that won't happen.

peckpeck:
@jooooooon yes, I was just replying to @ncharles who wanted to target it to 3.2

jooooooon:
Oh yes, sorry @peckpeck, somehow I missed the entire conversation. That's ok then.

PR #945, opened 2016-05-24
Fixes #8366: 8365 broke agent build

amousset:
https://www.rudder-project.org/redmine/issues/8366


amousset:
Fixed while upmerging #946

PR #948, opened 2016-05-25
Fixes #8369: Rudder jetty init script broken

amousset:
https://www.rudder-project.org/redmine/issues/8369


amousset:
The version check is already done in the init script.

jooooooon:
This should also be fixed in 2.11 in :

`rudder-jetty/SOURCES/rudder-jetty.default:27:JAVA_MAJOR_VERSION=`${JAVA} -version 2>&1 | grep "^java version" | sed 's/java version "\([0-9]\+\.[0-9]\)\+\..*"/\1/'`
`

jooooooon:
As for the duplicate version check, I propose instead to remove rudder-jetty/SOURCES/patches/jetty/jetty-init-check-java-version.patch, and keep the one in rudder-jetty/SOURCES/rudder-jetty.default because it's much easier to modify scripts than patches, and the error message is better there

amousset:
@jooooooon Replaced by #950 for java detection, opened https://www.rudder-project.org/redmine/issues/8373 to remove the check from the init script.

PR #949, opened 2016-05-25
Fixes #8371: Spurious message \"grep: /etc/sysconfig/apache2: No such file or directory\" in RPM install for rudder-webapp

jooooooon:
https://www.rudder-project.org/redmine/issues/8371


peckpeck:
Good, merging

PR #950, opened 2016-05-25
Fixes #8369: Rudder jetty init script broken on CentOS 7

amousset:
https://www.rudder-project.org/redmine/issues/8369


jooooooon:
Replaces #948

PR #952, opened 2016-05-25
Fixes #8372: Several systemd warnings in install on RHEL/CentOS 7  bugs in original resolution and missing changes

jooooooon:
https://www.rudder-project.org/redmine/issues/8372

### DIFF
--- a/rudder-inventory-endpoint/SPECS/rudder-inventory-endpoint.spec
+++ b/rudder-inventory-endpoint/SPECS/rudder-inventory-endpoint.spec
@@ -173,8 +173,14 @@ echo "INFO: Launching script to check if a migration is needed"
 %{rudderdir}/bin/rudder-inventory-endpoint-upgrade
 echo "INFO: End of migration script"
 
-echo "Restarting syslogd ..."
-service %{syslogservicename} restart
+echo -n "INFO: Restarting syslogd ..."
+%if 0%{?rhel} < 7
+service %{syslogservicename} restart > /dev/null && echo " Done"
+%endif
+%if 0%{?rhel} >= 7
+/bin/systemctl restart  %{syslogservicename}.service && echo " Done"
+%endif
+
 
 #=================================================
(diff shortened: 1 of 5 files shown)


rudder-inventory-endpoint/SPECS/rudder-inventory-endpoint.spec
```diff
@@ -173,8 +173,14 @@ echo "INFO: Launching script to check if a migration is needed"
...
 echo "INFO: End of migration script"
 
-echo "Restarting syslogd ..."
-service %{syslogservicename} restart
+echo -n "INFO: Restarting syslogd ..."
+%if 0%{?rhel} < 7
+service %{syslogservicename} restart > /dev/null && echo " Done"
+%endif
+%if 0%{?rhel} >= 7
+/bin/systemctl restart  rsyslog.service && echo " Done"
```
peckpeck: You could use %{syslogservicename}.service instead

rudder-reports/SPECS/rudder-reports.spec
```diff
@@ -117,14 +117,26 @@ else
 fi
 
 # Check if PostgreSQL is started
-service ${POSTGRESQL_SERVICE_NAME} status > /dev/null
+%if 0%{?rhel} < 7
+CHECK_COMMAND="service ${POSTGRESQL_SERVICE_NAME} status"
```
peckpeck: Why do you do differently here and use a variable instead of a direct call ?

jooooooon:
Commit modified

rudder-inventory-endpoint/SPECS/rudder-inventory-endpoint.spec
jooooooon: Good catch, fixing

jooooooon:
Commit modified

peckpeck:
Good, merging

PR #954, opened 2016-05-25
Fixes #8388: rudder-jetty build fails when applying a removed patch

VinceMacBuche:
https://www.rudder-project.org/redmine/issues/8388


jooooooon:
Good catch.

PR #957, opened 2016-06-02
Fixes #8424: When updating runs, hooks should really be async

ncharles:
https://www.rudder-project.org/redmine/issues/8424


ncharles:
wrong branch and commit name - closing

PR #961, opened 2016-06-03
Fixes #8447: Add some doc to the #8446 patch

amousset:
https://www.rudder-project.org/redmine/issues/8447


peckpeck:
OK, merging

PR #967, opened 2016-06-08
Fixes #8491: The new makefile doesn't work with old make

peckpeck:
https://www.rudder-project.org/redmine/issues/8491


amousset:
Ok, merging.

PR #968, opened 2016-06-08
Fixes #8493: pcre is used instead on openssl in the new Makefile

peckpeck:
https://www.rudder-project.org/redmine/issues/8493


amousset:
Ok, merging.

PR #969, opened 2016-06-08
Fixes #8495: Old OS don't support the --retry-connrefused on wget

peckpeck:
https://www.rudder-project.org/redmine/issues/8495


amousset:
Ok, merging.

PR #970, opened 2016-06-08
Fixes #8496: Remove old distribution specific patch files in package 3.3

peckpeck:
https://www.rudder-project.org/redmine/issues/8496


amousset:
Ok, merging.

PR #971, opened 2016-06-09
Fixes #8497: References to the desdir in the destdir itself in rudder-agent (in perl files)

amousset:
https://www.rudder-project.org/redmine/issues/8497


peckpeck:
OK, merging

PR #972, opened 2016-06-09
Fixes #8499: Removed files are still in the source spec file on master

peckpeck:
https://www.rudder-project.org/redmine/issues/8499


amousset:
Ok, merging

PR #973, opened 2016-06-09
Fixes #8502: RPM package include some file twice

peckpeck:
https://www.rudder-project.org/redmine/issues/8502


amousset:
Ok, merging

PR #974, opened 2016-06-09
Fixes #8504: do not package ld.so.conf.d in RHEL3

peckpeck:
https://www.rudder-project.org/redmine/issues/8504


amousset:
Ok, merging.

PR #975, opened 2016-06-09
Fixes #8505: Conffile are not the same on debian and rehdat

peckpeck:
https://www.rudder-project.org/redmine/issues/8505


rudder-agent/SPECS/rudder-agent.spec
```diff
@@ -768,6 +768,12 @@ rm -f %{_builddir}/file.list.%{name}
 %config(noreplace) /etc/ld.so.conf.d/rudder.conf
 %endif
 
+%config(noreplace) /opt/rudder/etc/uuid.hive
+%config(noreplace) /etc/cron.d/rudder-agent
+%config(noreplace) /etc/init.d/rudder-agent
+%config(noreplace) /etc/default/rudder-agent
+%config(noreplace) /etc/profile.d/rudder-agent.sh
```
jooooooon: I don't think we want to mark these as config files on RPM based systems.

The reason they are on deb-based systems is that dpkg will ask the user what to do if a config file is modified, and usually a user will just replace it if he doesn't know what's going on.

However, rpm does not ask questions, and silently places a <file>.rpmnew next to the target file. This results in upgrades not actually upgrading important files - in the list above, the cron and init files (at least) are very important, and need to be updated. This is why we removed them as config files, to make sure they were overwritten.

peckpeck: Then we could have a simple conffile (without noreplace) and rpm would replace the file but keep the old one in a .rpmsave file

peckpeck: Why was /etc/ld.so.conf.d/rudder.conf a %config(noreplace) file ?

jooooooon: No idea :)

jooooooon: Simple conffile sounds good to me.

peckpeck:
Commit modified

rudder-agent/SPECS/rudder-agent.spec
amousset: Shouldn't we set /etc/default/rudder-agent as config(noreplace), as it is only there to let the user override variables in the init script?

peckpeck:
Commit modified

rudder-agent/SPECS/rudder-agent.spec
```diff
@@ -765,9 +765,15 @@ rm -f %{_builddir}/file.list.%{name}
 %dir %{rudderlogdir}/install
 
 %if 0%{?rhel} != 3 && "%{?_os}" != "aix"
-%config(noreplace) /etc/ld.so.conf.d/rudder.conf
+%config /etc/ld.so.conf.d/rudder.conf
 %endif
 
+%config /opt/rudder/etc/uuid.hive
```
jooooooon: I think this should be "noreplace" otherwise we will replace user's uuid.hive files with the empty default on each upgrade

peckpeck:
Commit modified

rudder-agent/SPECS/rudder-agent.spec
```diff
@@ -765,9 +765,15 @@ rm -f %{_builddir}/file.list.%{name}
 %dir %{rudderlogdir}/install
 
 %if 0%{?rhel} != 3 && "%{?_os}" != "aix"
-%config(noreplace) /etc/ld.so.conf.d/rudder.conf
+%config /etc/ld.so.conf.d/rudder.conf
 %endif
 
+%config /etc/cron.d/rudder-agent
+%config /etc/profile.d/rudder-agent.sh
+%config /etc/init.d/rudder-agent
```
jooooooon: Is this really to be considered a "config" file?

peckpeck:
Commit modified

rudder-agent/SPECS/rudder-agent.spec
```diff
@@ -765,9 +765,14 @@ rm -f %{_builddir}/file.list.%{name}
 %dir %{rudderlogdir}/install
 
 %if 0%{?rhel} != 3 && "%{?_os}" != "aix"
-%config(noreplace) /etc/ld.so.conf.d/rudder.conf
+%config /etc/ld.so.conf.d/rudder.conf
```
jooooooon: This seems important, let's keep that.

rudder-agent/SPECS/rudder-agent.spec
```diff
@@ -765,9 +765,14 @@ rm -f %{_builddir}/file.list.%{name}
...
 %endif
 
+%config /etc/cron.d/rudder-agent
+%config /etc/profile.d/rudder-agent.sh
```
jooooooon: I don't see why we are changing these in a released versoin - does this fix a bug? I'm fine with doing it in master just not on 2.11 unless we have a reason.

rudder-agent/SPECS/rudder-agent.spec
```diff
@@ -765,9 +765,14 @@ rm -f %{_builddir}/file.list.%{name}
...
+%config(noreplace) /opt/rudder/etc/uuid.hive
+%config(noreplace) /etc/default/rudder-agent
```
jooooooon: OK.

peckpeck:
Superseded by https://github.com/Normation/rudder-packages/pull/989

PR #977, opened 2016-06-09
Fixes #8507: RPM package doesn't include /etc files

peckpeck:
https://www.rudder-project.org/redmine/issues/8507


amousset:
I could not find where etc/default/rudder-agent gets installed in the DESTDIR.

peckpeck:
You're right, here it is http://www.rudder-project.org/redmine/issues/8508

amousset:
Ok, merging.

PR #980, opened 2016-06-14
Fixes #8526: Error with java build dependency on SLES11

amousset:
https://www.rudder-project.org/redmine/issues/8526


amousset:
Commit modified

PR #981, opened 2016-06-14
Fixes #8530: Error with java build dependency on SLES11

amousset:
https://www.rudder-project.org/redmine/issues/8530


amousset:
Commit modified

PR #982, opened 2016-06-14
Fixes #8533: cpanm doesn't build on rhel 3 64 bits

peckpeck:
https://www.rudder-project.org/redmine/issues/8533


amousset:
Ok, merging.

PR #983, opened 2016-06-16
Fixes #8500: make rudder-perl able to call either rudder's perl or the local perl

peckpeck:
https://www.rudder-project.org/redmine/issues/8500

### DIFF
--- a/rudder-agent/SOURCES/rudder-perl
+++ b/rudder-agent/SOURCES/rudder-perl
@@ -1,5 +1,10 @@
 #!/bin/sh
 #
-# This script runs a Perl script using Rudder's Perl environment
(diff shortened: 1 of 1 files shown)


jooooooon:
With the very similar script "run-inventory" that looks for a valid FusionInventory agent to run, we found it very useful for later debugging to add a line saying which command is being used: https://github.com/Normation/rudder-packages/blob/master/rudder-agent/SOURCES/run-inventory#L27

Could we add that here too?

peckpeck:
Perl can be used by a lots of things, the output of a perl command can be important, I think we should not touch it

jooooooon:
Hmm, good point.

PR #984, opened 2016-06-16
Fixes #8454: On SLES (at least), updating rudder-server-root does not update dependencies

peckpeck:
https://www.rudder-project.org/redmine/issues/8454


fanf:
@peckpeck will check that there is no missing dependencies before merging.

peckpeck:
Commit modified

PR #988, opened 2016-06-24
Fixes #8602: 

amousset:
https://www.rudder-project.org/redmine/issues/8602

### DIFF
--- a/ncf-api-virtualenv/SOURCES/ncf-api-virtualenv.conf
+++ b/ncf-api-virtualenv/SOURCES/ncf-api-virtualenv.conf
@@ -4,7 +4,7 @@
 Alias /ncf-builder /usr/share/ncf/builder
 
 ## Set up a WSGI serving process named ncf_api_flask_app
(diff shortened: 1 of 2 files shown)


ncf-api-virtualenv/SOURCES/ncf-api-virtualenv.conf
```diff
@@ -4,7 +4,7 @@
 Alias /ncf-builder /usr/share/ncf/builder
 
 ## Set up a WSGI serving process named ncf_api_flask_app
-WSGIDaemonProcess ncf_api_flask_app threads=5 user=ncf-api-venv
+WSGIDaemonProcess ncf_api_flask_app threads=5 user=ncf-api-venv umask=0007
```
ncharles: promises say 600, but you set 660 here, why the difference ?

amousset: The files written by the builder are `ncf-api-venv:rudder`, whereas the content of `ncf/local` is `root:root`.

PR #992, opened 2016-07-08
Fixes #8658: Update openssl to 1.0.1t on 2.11 branch

amousset:
https://www.rudder-project.org/redmine/issues/8658


amousset:
We also need the source in http://www.normation.com/tarball/openssl/openssl-1.0.1t.tar.gz.

amousset:
OK, merging this PR

PR #1000, opened 2016-07-26
Fixes #8760

FlorianHeigl:
very limited change to debian flavours
the tests will be sequential like this.
before that it'd try to find ubuntu specific things on debian and give an error about it each time the script is called.


peckpeck:
Thank you

PR #1001, opened 2016-08-03
Fixes #7937: Check and/or enhance ncf usage in initial promises

peckpeck:
https://www.rudder-project.org/redmine/issues/7937

### DIFF
--- a/rudder-agent/SOURCES/Makefile
+++ b/rudder-agent/SOURCES/Makefile
@@ -190,6 +190,7 @@ endif
 	$(INSTALL) -m 755 uuidgen $(INSTALL_DIR)/bin/
 
 	cp -r initial-promises/ $(INSTALL_DIR)/share/
+	cp -r initial-ncf/ $(INSTALL_DIR)/share/
 
 # clean 
 clean: localclean
@@ -207,7 +208,7 @@ mrproper: veryclean
 veryclean:
 	rm -f rudder-agent.cron
 	rm -f ./rudder-sources.tar.bz2
-	rm -rf initial-promises rudder-agent-utilities fusioninventory-agent
+	rm -rf initial-promises initial-ncf rudder-agent-utilities fusioninventory-agent
 	rm -rf cfengine-source lmdb-source openssl-source perl-$(PERL_VERSION)
 
 ##########################################
@@ -276,9 +277,15 @@ rudder.8.gz:
 	$(WGET) http://www.rudder-project.org/rudder-doc-${RUDDER_MAJOR_VERSION}/rudder.8
 	gzip -f rudder.8
 
-initial-promises: ./rudder-sources
+initial-promises: ./rudder-sources initial-ncf
 	rm -rf ./initial-promises/
 	cp -r ./rudder-sources/rudder-techniques/initial-promises/node-server/ ./initial-promises
+
+initial-ncf: ./rudder-sources
+	rm -rf ./initial-ncf
+	cp -r ./rudder-sources/ncf/tree ./initial-ncf
+	# retreive the rudder specific ncf.conf
+	cp ./initial-promises/node-server/distributePolicy/ncf/rudder-ncf-conf ./initial-ncf/ncf.conf
 
 rudder-agent-utilities: rudder-sources
 	rm -rf ./rudder-agent-utilities/
--- a/rudder-agent/SOURCES/check-rudder-agent
+++ b/rudder-agent/SOURCES/check-rudder-agent
@@ -103,9 +103,9 @@ check_and_fix_cfengine_processes() {
     echo " Done"
   fi
 
-  # Get the value of rudder-agent run interval from file /var/rudder/cfengine-community/inputs/run_interval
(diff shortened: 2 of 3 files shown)


rudder-agent/SOURCES/rudder-agent-postinst
```diff
@@ -73,6 +73,8 @@ if [ ! -f /var/rudder/cfengine-community/inputs/promises.cf ] || ! /var/rudder/c
 then
   rm -rf /var/rudder/cfengine-community/inputs/* || true
   cp -r /opt/rudder/share/initial-promises/* /var/rudder/cfengine-community/inputs/
+  rm -rf /var/rudder/ncf/common/*
+  cp -r /opt/rudder/share/initial-ncf/* /var/rudder/ncf/common/
 fi
```
jooooooon: I think this whole block of code should live in check-rudder-agent (or soon to be "rudder agent check"). The command is already called at the end of this script (the last thing that happens) and it should be doing this reset anyway.

However, that would likely involve changing the order below so that the agent start command comes after the check.

rudder-agent/SOURCES/check-rudder-agent
```diff
@@ -103,9 +103,9 @@ check_and_fix_cfengine_processes() {
     echo " Done"
   fi
 
-  # Get the value of rudder-agent run interval from file /var/rudder/cfengine-community/inputs/run_interval
-  if [ -f "${CFE_DIR}/inputs/run_interval" ]; then
-    RUN_INTERVAL=`cat "${CFE_DIR}/inputs/run_interval"`
+  # Get the value of rudder-agent run interval from file /var/rudder/cfengine-community/inputs/run-interval
+  if [ -f "${CFE_DIR}/inputs/run-interval" ]; then
+    RUN_INTERVAL=`cat "${CFE_DIR}/inputs/run-interval"`
```
jooooooon: Is this supposed to be here? It seems unrelated.

rudder-agent/SOURCES/Makefile
```diff
@@ -276,10 +277,14 @@ rudder.8.gz:
...
 	gzip -f rudder.8
 
-initial-promises: ./rudder-sources
+initial-promises: ./rudder-sources initial-ncf
 	rm -rf ./initial-promises/
 	cp -r ./rudder-sources/rudder-techniques/initial-promises/node-server/ ./initial-promises
 
+initial-ncf: ./rudder-sources
+	rm -rf ./initial-ncf
+	cp -r ./rudder-sources/ncf/tree ./initial-ncf
```
jooooooon: For initial-ncf to work, we'll need an initial ncf.conf file too, that should contain the configuration to use log_rudder, and the 5309 CFEngine port (and whatever else we change in there).

rudder-agent/SOURCES/check-rudder-agent
peckpeck: The name has been changed in the initial promises merge PRs, since this one is the continuation of those PRs i've chosen to put it here.

rudder-agent/SOURCES/rudder-agent-postinst
peckpeck: We could put the agent startup code below in check-rudder-agent too.
The only difference is that would not have an install specific message anymore

peckpeck:
Commit modified
Add rudder specific ncf.conf

rudder-agent/SOURCES/check-rudder-agent
jooooooon: OK.

rudder-agent/SOURCES/rudder-agent-postinst
jooooooon: In the case of a "normal" start (agent not disabled + not first install) there is no specific install message, so that's fine by me. We do need to keep the install specific messages when agent is disabled or not yet configured, here, though.

rudder-agent/SOURCES/Makefile
```diff
@@ -276,10 +277,16 @@ rudder.8.gz:
...
+	# retreive the rudder specific ncf.conf
+	cp ./initial-promises/node-server/distributePolicy/ncf/rudder-ncf-conf ./initial-ncf/ncf.conf
```
peckpeck: Added here

rudder-agent/SOURCES/check-rudder-agent
peckpeck: I will do this in another PR to make it easier to read/merge.
See http://www.rudder-project.org/redmine/issues/8814

peckpeck: Sorry, this comment belongs to the thread below, about moving things to check-rudder-agent

rudder-agent/SOURCES/Makefile
jooooooon: :+1:

rudder-agent/SOURCES/check-rudder-agent
jooooooon: OK

PR #1002, opened 2016-08-03
Fixes #8807: Move key permission checks from postinst to check-rudder-agent

peckpeck:
https://www.rudder-project.org/redmine/issues/8807

### DIFF
--- a/rudder-agent/SOURCES/check-rudder-agent
+++ b/rudder-agent/SOURCES/check-rudder-agent
@@ -60,6 +60,16 @@ clean_cf_lock_files() {
   rm -f ${CFE_DIR}/state/cf_lock.${CFENGINE_DB_EXT}.lock
 }
 
+check_and_fix_cfengine_keys() {
+  # Making sure that the permissions on the CFEngine key directory are correct...
+  if [ -d ${CFE_DIR}/ppkeys ]; then
+    chmod 700 ${CFE_DIR}/ppkeys
+    if [ `ls ${CFE_DIR}/ppkeys | wc -l` -gt 0 ]; then
+      chmod 600 ${CFE_DIR}/ppkeys/*
+    fi
+  fi
+}
+
 check_and_fix_cfengine_processes() {
 
   # Detect the correct ps tool to use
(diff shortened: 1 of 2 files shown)


peckpeck:
Commit modified

rudder-agent/SOURCES/check-rudder-agent
```diff
@@ -60,6 +60,25 @@ clean_cf_lock_files() {
   rm -f ${CFE_DIR}/state/cf_lock.${CFENGINE_DB_EXT}.lock
 }
 
+fix_perm() {
+  perm="$1"
+  file="$2"
+  if [ `stat -c %a "${file}"` != "${perm}" ]; then
```
jooooooon: This stat command won't work on AIX (there is no stat command)

peckpeck: We could just do the chmod as before but the goal was to avoid doing it unnecessarily.
Do you think it is a worthy goal ?

jooooooon: I think it's probably not worth it, because I'm pretty sure the chmod command itself already does a similar test to avoid unnecessary writes.

peckpeck:
Commit modified
Remove stat call

rudder-agent/SOURCES/check-rudder-agent
```diff
@@ -267,6 +277,7 @@ then
   CFE_BIN_DIR="${CFE_DIR}/bin"
   LAST_UPDATE_FILE="${CFE_DIR}/last_successful_inputs_update"
   NEED_INIT_FILES="true"
+  check_and_fix_cfengine_keys
```
jooooooon: You only added this in the "Once for rudder's cfengine" section, shouldn't we also do it in the "Once for enterprise cfengine" section below? I know we never have before, but if it's useful for one CFEngine it must surely be for the other too.

peckpeck: indeed

peckpeck:
Commit modified

PR #1003, opened 2016-08-03
Fixes #8808: Use rudder agent commands when possible in postinst

peckpeck:
https://www.rudder-project.org/redmine/issues/8808

PR #1004, opened 2016-08-03
Fixes #8806: Move check-rudder-agent to rudder-agent repo

peckpeck:
https://www.rudder-project.org/redmine/issues/8806


rudder-agent/SOURCES/check-rudder-agent
```diff
@@ -1,291 +1,3 @@
...
-fi
-
-# The following files are not present on AIX systems
-if [ "${NEED_INIT_FILES}" = "true" ] && [ "${OS_FAMILY}" != "AIX" ] || [ "${OS_FAMILY}" = "SunOS" ]; then
-  check_and_fix_specific_rudder_agent_file /etc/init.d/rudder-agent rudder-agent.init "Please reinstall the rudder-agent package." "mandatory"
-  check_and_fix_specific_rudder_agent_file /etc/default/rudder-agent rudder-agent.default "Please reinstall the rudder-agent package." "mandatory"
-  check_and_fix_specific_rudder_agent_file /etc/cron.d/rudder-agent rudder-agent.cron "Please reinstall the rudder-agent package." "mandatory"
-fi
+# this file content has been moved
+exec /opt/rudder/share/commands/agent-check
```
jooooooon: Same question as in #1003, why not call "rudder agent check" to use the main wrapper?

jooooooon:
This is not directly related to this PR, but I think we should also change the cron file we install to directly call "rudder agent check" instead of the check-rudder-agent script, which will soon become obsolete.

peckpeck:
It's true that we should do it, but it must be done twice to work, once in the packaging and once in the system technique, we can do one here.

peckpeck:
Commit modified

peckpeck:
The cron part will be done in  #8808 and in #8813

peckpeck:
PR rebased

PR #1006, opened 2016-08-05
Fixes #8822: Missing BuildRequires for rudder-agent : pam-devel

amousset:
https://www.rudder-project.org/redmine/issues/8822

### DIFF
--- a/rudder-agent/SPECS/rudder-agent.spec
+++ b/rudder-agent/SPECS/rudder-agent.spec
@@ -87,7 +87,7 @@ Patch1: fix-missing-headers
 BuildRoot: %{_tmppath}/%{name}-%{version}-%{release}-root-%(%{__id_u} -n)
 
 # Generic requirements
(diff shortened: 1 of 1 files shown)


amousset:
Commit modified

amousset:
ping @jooooooon

jooooooon:
Ah, sorry, I should have answered this earlier. I just don't understand this - we have been building the agent on 2.11 for years, why would we need a new BuildRequires now?

The ticket doesn't help me understand the context either :)

amousset:
It is installed on every builder because of https://github.com/Normation/rudder-packages/blob/master/rudder-agent/SOURCES/.dependencies, and also in the requirements for the deb package, but missing from the spec file.

jooooooon:
Ah, I see! Thanks for explaining. For future reference, this is *exactly* what should be written in the ticket itself.

amousset:
OK, merging this PR

PR #1011, opened 2016-08-11
Fixes #8835: Broken 3.3 builds - missing file

amousset:
https://www.rudder-project.org/redmine/issues/8835


amousset:
Commit modified

PR #1012, opened 2016-08-17
Fixes #8855: 8822 breaks rudder-agent-thin patch

amousset:
https://www.rudder-project.org/redmine/issues/8855


amousset:
OK, merging this PR

PR #1015, opened 2016-08-18
Fixes #8868:  Failing CFEngine unit test on recent OSes - Remove patch in master

amousset:
https://www.rudder-project.org/redmine/issues/8868


amousset:
Commit modified

PR #1017, opened 2016-08-25
Fixes #8893: Do not depend on pam-devel on AIX

amousset:
https://www.rudder-project.org/redmine/issues/8893


amousset:
OK, merging this PR

PR #1019, opened 2016-08-25
Fixes #8895: Change rudder-agent-thin patches according to 8893

amousset:
https://www.rudder-project.org/redmine/issues/8895


amousset:
OK, merging this PR

PR #1023, opened 2016-08-30
Fixes #8876: The file rsyslog.d/rudder.conf might not be embedded in rudder-reports (db on a separate server)

amousset:
https://www.rudder-project.org/redmine/issues/8876


amousset:
OK, merging this PR

PR #1024, opened 2016-08-30
Fixes #8837: Missing makeuuid on RPM systems

amousset:
https://www.rudder-project.org/redmine/issues/8837

### DIFF
--- a/rudder-agent/SOURCES/uuidgen
+++ b/rudder-agent/SOURCES/uuidgen
@@ -1,6 +1,7 @@
 #!/bin/sh
 
 # uuidgen doesn't exist on AIX, so provide a compatible shell function
+# This is not compatible with Linux systems.
 
 if [ -f /etc/ncs/uuid_gen ]
 then
--- a/rudder-agent/SPECS/rudder-agent.spec
+++ b/rudder-agent/SPECS/rudder-agent.spec
@@ -47,6 +47,9 @@
 # Same goes for the use of the local PCRE install vs. a bundled one
 %define use_system_pcre true
 
+# Should we embeb a uuidgen equivalent (i.e. for AIX)
+%define use_system_uuidgen true
+
 # Perl and fusion
 %if "%{real_name}" == "rudder-agent"
 %define use_system_fusion false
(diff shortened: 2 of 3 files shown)


rudder-agent/SPECS/rudder-agent.spec
```diff
@@ -47,6 +47,9 @@
 # Same goes for the use of the local PCRE install vs. a bundled one
 %define use_system_pcre true
 
+# Should we embeb a uuidgen equivalent (i.e. for AIX)
```
jooooooon: embed*

jooooooon: And this is the wrong way round - it should say "Should we rely on system uuidgen (if false, provide our own equivalent, i.e. for AIX)"

rudder-agent/SOURCES/uuidgen
```diff
@@ -1,6 +1,7 @@
 #!/bin/sh
 
 # uuidgen doesn't exist on AIX, so provide a compatible shell function
+# This is not compatible with Linux systems.
```
jooooooon: Why not also add here a `if type uuidgen > /dev/null 2> /dev/null` protection, just in case, as we discussed?

rudder-agent/SOURCES/uuidgen
```diff
@@ -1,6 +1,7 @@
...
```
amousset: Not sure how we should handle the case where there is a uuidgen binary on the system. 
* When this script is called, it is added to the path, so we would need to check the path of the script and see if there is another available uuidgen...
* We could also change agent-check to stop adding /opt/rudder/bin to the path and choose our uuidgen only if it cannot find one on the system.
* Finally, I don't really like calling it uuidgen as it is not a drop-in replacement (it does not support any options).

We could move all this logic to rudder-agent common lib and have a standard internal uuid generation with platform detection.

rudder-agent/SOURCES/uuidgen
amousset: @jooooooon Because I wasn't sure how to do it, and even if we should really do it this way, see my comment below.

rudder-agent/SOURCES/uuidgen
peckpeck: I would prefer putting all the logic in a single place (this script, but it can be renamed) and  call it from different places, we currently have rudder agent check, rudder agent factory-reset, the postinstall script, and probably others.

So i would prefer this script to handle all the possible cases.

amousset:
Replaced by #1039.

PR #1030, opened 2016-09-06
Fixes #8928:  Agent sometimes fail to apply package actions because of global lock - 3.2 version

amousset:
https://www.rudder-project.org/redmine/issues/8928


amousset:
The fix is exactly the same, we just add the tests back.

PR #1033, opened 2016-09-07
Fixes #8850: We need a run with the initial promises before being able to update the promises

peckpeck:
https://www.rudder-project.org/redmine/issues/8850


peckpeck:
OK, merging this PR

PR #1034, opened 2016-09-07
Fixes #8968: Postinst doesn't create the directories it needs

peckpeck:
https://www.rudder-project.org/redmine/issues/8968


peckpeck:
OK, merging this PR

PR #1042, opened 2016-09-12
Fixes #9010: Add an automatic network detection in rudder-init

peckpeck:
https://www.rudder-project.org/redmine/issues/9010

### DIFF
--- a/rudder-webapp/SOURCES/rudder-init
+++ b/rudder-webapp/SOURCES/rudder-init
@@ -37,6 +37,7 @@ RUDDER_OPT="/opt/rudder"
 RUDDER_VAR="/var/rudder"
 
 RUDDER_ROLES_FILE="${RUDDER_VAR}/cfengine-community/inputs/rudder-server-roles.conf"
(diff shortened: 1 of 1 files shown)


peckpeck:
doesn't work, do not merge

rudder-webapp/SOURCES/rudder-init
```diff
@@ -192,6 +192,21 @@ function reset_policy_server() {
...
+    # list all ipv4 networks matching each non loopback interface
+    for i in $(LANG=C ip -family inet -oneline address | grep -v " lo " | sed -e 's/.*inet \([0-9.\/]\+\) .*/\1/')
+    do
+      ALLOWEDNETWORK[${#ALLOWEDNETWORK[*]}]="${i}"
+    done
+  else
+    # just use the provided network
+    ALLOWEDNETWORK[${#ALLOWEDNETWORK[*]}]="${network}"
+  fi
+}
```
jooooooon: As discussed today, we should add a check for an environment variable and a flag file that restricts this auto function to use /32 IP addresses (we don't want to have NO allowed networks because this would break several things).

peckpeck:
Commit modified

peckpeck:
Commit modified

PR #1044, opened 2016-09-12
Fixes #9000: Maven does not honor PROXY settings

bmwjanos:
Generate Maven's proxy setting from template according to PROXY
environment variable being set

### DIFF
--- a/rudder-webapp/SOURCES/Makefile
+++ b/rudder-webapp/SOURCES/Makefile
@@ -19,6 +19,18 @@
 MAVEN_RELEASE = 3.0.4
 WGET := $(if $(PROXY), http_proxy=$(PROXY) ftp_proxy=$(PROXY)) /usr/bin/wget -q
 
+MAVEN_PROXY := $(if $(PROXY), $(shell echo "$(PROXY)" |  perl -ne ' \
+    if (m|(https?)://(?:([^:@]+)(?::([^@]+))?\@)?([^:/]+)(?::([0-9]+))?|) { \
+        chomp;
+        my $$proto = "<protocol>$$1</protocol>"; \
+        my $$user = defined($$2) ? "<username>$$2</username>" : ""; \
+        my $$pass = defined($$3) ? "<password><![CDATA[$$3]]></password>" : ""; \
+        my $$host = "<host>$$4</host>";  \
+        my $$id   = "<id>$$4</id>"; \
+        my $$port = defined($5) ? "<port>$$5</port>" : ""; \
+        print "<proxies> <proxy> <active>true</active> $$id $$proto $$user $$pass $$host $$port </proxy> </proxies>"; \
+    }'))
+
 .DEFAULT_GOAL := localdepends
 
 RUDDER_VERSION_TO_PACKAGE = <put Rudder version or version-snapshot here>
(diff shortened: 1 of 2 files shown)


rudder-webapp/SOURCES/Makefile
```diff
@@ -19,6 +19,17 @@
 MAVEN_RELEASE = 3.0.4
 WGET := $(if $(PROXY), http_proxy=$(PROXY) ftp_proxy=$(PROXY)) /usr/bin/wget -q
 
+MAVEN_PROXY := $(if $(PROXY), $(shell echo "$(PROXY)" |  perl -ne ' \
+    if (m|(https?)://(?:([^:@]+)(?::([^@]+))?\@)?([^:/]+)(?::([0-9]+))?|) { \
```
peckpeck: '(?:' means do not remember this field, so the $x variables where x>=2 will be empty

rudder-webapp/SOURCES/Makefile
```diff
@@ -19,6 +19,17 @@
...
+        my $$proto = "<protocol>$$1</protocol>"; \
+        my $$user = defined($$2) ? "<username>$$2</username>" : ""; \
+        my $$pass = defined($$3) ? "<password><![CDATA[$$3]]></password>" : ""; \
+        my $$host = "<host>$$4</host>";  \
+        my $$id   = "<id>$$4</id>"; \
+        my $$port = defined($5) ? "<port>$$5</port>" : ""; \
+        print "<proxies> <active>true</active> $$id $$proto $$user $$pass $$host $$port </proxies>"; \
```
peckpeck: you are missing a <proxy> inside your <proxies>

rudder-webapp/SOURCES/Makefile
bmwjanos: What is the `$x` variable you meant?

peckpeck: I mean $1 $2 $3 $4 $5 below

bmwjanos: $1 and $2 are true - inside a (?:, but they are not (?: themselves, so they are actually undefined.
so I have basically a match set of `<$1:proto>://<$2:user>:<$3:pass>@<$4:host>:<$5:port>`, from which $2/$3/$5 might be missing, and therefore undefined.
You can try this by this short test script:
```
#!/bin/bash
for PROXY in \
        'http://localhost' \
        'http://localhost:1234' \
        'https://user@localhost' \
        'https://user@sub.proxy.fqdn:12345' \
        'https://user:pass@proxy.fqdn' \
        'https://user:pass@proxy.fqdn:port'; do
 echo "====> $PROXY"
 echo "$PROXY" |  perl -ne '
     chomp;
     if (m|(https?)://(?:([^:@]+)(?::([^@]+))?\@)?([^:/]+)(?::([0-9]+))?|) {
         my $proto = "<protocol>$1</protocol>";
         my $user = defined($2) ? "<username>$2</username>" : "";
         my $pass = defined($3) ? "<password><![CDATA[$3]]></password>" : "";
         my $host = "<host>$4</host>";
         my $id   = "<id>$4</id>";
         my $port = defined($5) ? "<port>$5</port>" : "";
         print "<proxies> <proxy> <active>true</active> $id $proto $user $pass $host $port </proxy> </proxies>\n";
     }';
 echo "";
done
```

PS: This already includes the suggested fixes and a chomp.

peckpeck: Ah I hadn't seen that you put regular parens inside (?: )

rudder-webapp/SOURCES/Makefile
bmwjanos: fixed that.

peckpeck:
Thanks

peckpeck:
Thanks

PR #1045, opened 2016-09-12
Fix #9002: Make downloading files in Makefile be done in two steps

bmwjanos:
Download to different file than required by build target, since even
a failing wget will create the file, and because of that the next
(probably correct) run will not retry the download.


peckpeck:
Good, thanks !

PR #1046, opened 2016-09-12
Fixes #8999: Added detection for SLES12 SP1+

bmwjanos:
SLES12 SP1 is not detected properly due to /etc/SuSE-release being
obsolete, and only /etc/os-release is present.

### DIFF
--- a/detect_os.sh
+++ b/detect_os.sh
@@ -37,6 +37,12 @@ elif [ -f /etc/SuSE-release ]; then
   export OS="SLES"
   export OSVERSION=$(cat /etc/SuSE-release | grep VERSION | cut -f2 -d '=' | sed 's/ //')
   export OSSP=$(cat /etc/SuSE-release | grep PATCHLEVEL | cut -f2 -d '=' | sed 's/ //')
+elif [ -f /etc/os-release ] && grep -q -s 'NAME="SLES"' /etc/os-release; then
+  # This file is present as of SLES12 SP1, and SuSE-release is obsoletes, so the logic above should
+  # still catch SLES12 GA before this becomes valid.
+  export OS="SLES"
+  export OSVERSION=$( eval "$( grep ^VERSION_ID /etc/os-release )"; echo "$VERSION_ID" | awk -F'.' '{ print $1 }' )
+  export OSSP=$( eval "$( grep ^VERSION_ID /etc/os-release )"; echo "$VERSION_ID" | awk -F'.' '{ print (NF==2) ? $2 : "0";}' )
 elif [ -f /etc/fedora-release ]; then
   export OS="FEDORA"
   export OSVERSION=$(cat /etc/fedora-release | sed 's/^.* release \([^\.][^\.]\).*$/\1/')


detect_os.sh
```diff
@@ -37,6 +37,12 @@ elif [ -f /etc/SuSE-release ]; then
   export OS="SLES"
   export OSVERSION=$(cat /etc/SuSE-release | grep VERSION | cut -f2 -d '=' | sed 's/ //')
   export OSSP=$(cat /etc/SuSE-release | grep PATCHLEVEL | cut -f2 -d '=' | sed 's/ //')
+elif [ -f /etc/os-release ] && grep -q -s 'NAME="SLES"' /etc/os-release; then
+  # This file is present as of SLES12 SP1, and SuSE-release is obsoletes, so the logic above should
+  # still catch SLES12 GA before this becomes valid.
+  export OS="SLES"
+  export OSVERSION=$( eval "$( grep ^VERSION_ID /etc/os-release )"; echo "$VERSION_ID" | cut -d. -f1 )
+  export OSSP=$( eval "$( grep ^VERSION_ID /etc/os-release )"; echo "$VERSION_ID" | cut -d. -f2 )
```
peckpeck: If there is no SP, this will not contain 0, as we would expect

bmwjanos: On SLES12 GA there is a SuSE-release file, so this elif is not executed but the case it catched by the previous elif in L36.
This only applies as of SLES12 SP1+, which actually *does* have an SP. Read the commends in L41+L42 :)

peckpeck: Yes, but that means we will have to patch again when there will be a sles 13

bmwjanos: In 2022? :) - ok modified it to be generic, and detect "0" if no SP is present.

detect_os.sh
```diff
@@ -37,6 +37,12 @@ elif [ -f /etc/SuSE-release ]; then
   export OS="SLES"
   export OSVERSION=$(cat /etc/SuSE-release | grep VERSION | cut -f2 -d '=' | sed 's/ //')
   export OSSP=$(cat /etc/SuSE-release | grep PATCHLEVEL | cut -f2 -d '=' | sed 's/ //')
+elif [ -f /etc/os-release ] && grep -q -s 'NAME="SLES"' /etc/os-release; then
+  # This file is present as of SLES12 SP1, and SuSE-release is obsoletes, so the logic above should
+  # still catch SLES12 GA before this becomes valid.
+  export OS="SLES"
+  export OSVERSION=$( eval "$( grep ^VERSION_ID /etc/os-release )"; echo "$VERSION_ID" | awk -F'.' '{ print $1 }' )
+  export OSSP=$( eval "$( grep ^VERSION_ID /etc/os-release )"; echo "$VERSION_ID" | awk -F'.' '{ print (NF==2) ? $2 : "0";' )
```
peckpeck: Good, except that you are missing a } at the en of your awk expression

peckpeck:
Thanks

PR #1052, opened 2016-09-13
Fixes #9030: We need a run with the initial promises before being able to update the promises

peckpeck:
https://www.rudder-project.org/redmine/issues/9030

### DIFF
--- a/rudder-agent/SOURCES/Makefile
+++ b/rudder-agent/SOURCES/Makefile
@@ -129,6 +129,8 @@ install: build $(INSTALL_DEPS) install-cfengine initial-promises initial-ncf rud
 	mkdir -p $(DESTDIR)/var/rudder/cfengine-community/inputs
 	mkdir -p $(DESTDIR)/var/rudder/tmp
 	mkdir -p $(DESTDIR)/var/rudder/tools
+	mkdir -p $(DESTDIR)/var/rudder/ncf/common
+	mkdir -p $(DESTDIR)/var/rudder/ncf/local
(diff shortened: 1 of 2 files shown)


rudder-agent/SOURCES/Makefile
```diff
@@ -129,6 +129,8 @@ install: build $(INSTALL_DEPS) install-cfengine initial-promises initial-ncf rud
 	mkdir -p $(DESTDIR)/var/rudder/cfengine-community/inputs
 	mkdir -p $(DESTDIR)/var/rudder/tmp
 	mkdir -p $(DESTDIR)/var/rudder/tools
+	-mkdir -p $(DESTDIR)/var/rudder/ncf/common
```
amousset: -mkdir -> mkdir

peckpeck:
Commit modified

rudder-agent/debian/dirs
```diff
@@ -7,7 +7,4 @@ var/rudder/cfengine-community/inputs
 var/rudder/cfengine-community/bin
 opt/rudder/etc
 var/rudder/tmp
-var/rudder/tools
```
amousset: Why removing it from dirs ?

peckpeck: because dirs is for directort creation and this creation is already done in the makefile

amousset: In this case we can remove almost all directories from dirs?

peckpeck: yes I think we should

amousset:
Closing as the solution will be different.

PR #1055, opened 2016-09-14
Fixes #8979: Fail while building perl modules on Ubuntu 16.04 for Rudder 3.1/3.2

amousset:
https://www.rudder-project.org/redmine/issues/8979


amousset:
OK, merging this PR

PR #1058, opened 2016-09-14
Fixes #9052:  Some KVM virtual machines considered as physical nodes (branch 4.0)

ncharles:
https://www.rudder-project.org/redmine/issues/9052


ncharles:
Commit modified

PR #1060, opened 2016-09-15
Fixes #9058: Broken RPM packageing of bash completion

amousset:
https://www.rudder-project.org/redmine/issues/9058


amousset:
OK, merging this PR

PR #1061, opened 2016-09-15
Fixes #9061:  Fail while building perl modules on Ubuntu 16.04 for Rudder 3.1/3.2 - same for rudder-agent-thin

amousset:
https://www.rudder-project.org/redmine/issues/9061


peckpeck:
I don't think adding a makefile would solve your problem

amousset:
Indeed...

PR #1062, opened 2016-09-15
Fixes #8921: The FusionInventory Rudder task does not work on Windows and will not return server roles

ncharles:
https://www.rudder-project.org/redmine/issues/8921


ncharles:
Commit modified

Kegeruneku:
The commit author needs to be amended :)

Kegeruneku:
Replaced by #1069

PR #1063, opened 2016-09-15
Fixes #9067: The FusionInventory Rudder task does not work on Windows and will not return server roles (branch 4.0)

ncharles:
https://www.rudder-project.org/redmine/issues/9067


Kegeruneku:
Replaced by #1070

PR #1085, opened 2016-09-27
Fixes #9042: Add a dependency on an ncf version with the new list-compatible-inputs

amousset:
https://www.rudder-project.org/redmine/issues/9042


amousset:
OK, merging this PR

PR #1088, opened 2016-09-28
Fixes #9156: Broken promises after 3.1.10 -> 3.1.14rc1 update

amousset:
https://www.rudder-project.org/redmine/issues/9156


ncharles:
Great work !

PR #1092, opened 2016-09-28
Fixes #9151: Add capabilities in inventory (4.0 style)

Kegeruneku:
https://www.rudder-project.org/redmine/issues/9151


Kegeruneku:
Was forced to update the previous patch too for things to apply correctly. This patch mess is hell :/

PR #1097, opened 2016-10-07
Fixes #9222: Wrong detection of rudder init parameters

peckpeck:
https://www.rudder-project.org/redmine/issues/9222

### DIFF
--- a/rudder-webapp/SOURCES/rudder-init
+++ b/rudder-webapp/SOURCES/rudder-init
@@ -195,6 +195,15 @@ function reset_policy_server() {
 
 function add_allowed_networks() {
   network="$1"
+  # check for invalid network definition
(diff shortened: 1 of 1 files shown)


rudder-webapp/SOURCES/rudder-init
```diff
@@ -195,6 +195,15 @@ function reset_policy_server() {
 
 function add_allowed_networks() {
   network="$1"
+  # check for invalid network definition
+  if [ $(echo -n "${network}" | sed 's|^[0-9]\+\.[0-9]\+\.[0-9]\+\.[0-9]\+/[0-9]\+$|OK|') != "OK" ]
```
amousset: Shouldn't we also support ipv6?

peckpeck: Good remark.
We should!
Do you know if it works in practice ?

peckpeck: tracked in a separate issue http://www.rudder-project.org/redmine/issues/9269

PR #1111, opened 2016-10-17
Fixes #9382: Change the Rudder version in the sources to be substituted in the packaging script

amousset:
https://www.rudder-project.org/redmine/issues/9382


rudder-webapp/SOURCES/Makefile
```diff
@@ -60,6 +61,9 @@ localdepends: ./rudder-sources ./maven/bin/mvn ./rudder-users.xml ./rudder-doc .
 	tar -xjf rudder-sources.tar.bz2
 	mv rudder-sources-*/ rudder-sources/
 
+	# Change rudder version displayed in the Web UI
+	sed -i "s%PACKAGING_RUDDER_VERSION%Rudder ${RUDDER_VERSION_TO_PACKAGE}%" rudder-sources/${HOMEPAGEPATH}
```
peckpeck: We should have a kind of tag delimiter like &lt; &gt; or &xx& 

Moreover we should replace the tag with the version only, because if we want to change the text we should only have to modify the original html file.

amousset:
Commit modified

amousset:
Commit modified

amousset:
Commit modified

amousset:
Commit modified

PR #1112, opened 2016-10-17
Fixes #9386: Only display major version on the login page

amousset:
https://www.rudder-project.org/redmine/issues/9386


amousset:
Commit modified

amousset:
Commit modified

PR #1117, opened 2016-10-19
Fixes #9412: Unexpected dependency on XML::TreePP on rudder-technique on Centos7

peckpeck:
https://www.rudder-project.org/redmine/issues/9412


peckpeck:
Commit modified

PR #1121, opened 2016-10-20
Fixes #9476: Add in packaging the migration scripts for database

ncharles:
https://www.rudder-project.org/redmine/issues/9476


rudder-webapp/SOURCES/rudder-upgrade
```diff
@@ -536,6 +537,15 @@ upgrade_database() {
     echo " Done"
   fi
 
+
+  # - 3.1.x and 3.2.x to 4.0.0: Migration DB schema to add archive table "archivedNodeConfigurations" and "ArchivedReportsExecution" 
+  RES=$(${PSQL} -t -d rudder -c "select count(*) from information_schema.columns where lower(table_name) = 'archivednodeconfigurations'")
```
peckpeck: Could you call it COUNT instead of RES ?
Because RES is an indication of a result code

fanf: All other command of the same kind have also a RES. So I for merging it as it, and changing all the RES to COUNT if wanted in a following release.

PR #1137, opened 2016-11-08
Fixes #6197: 

amousset:
https://www.rudder-project.org/redmine/issues/6197


peckpeck:
I understand why it was hard to find !

PR #1143, opened 2016-11-30
Fixes #9730: Package the new relay-api

peckpeck:
https://www.rudder-project.org/redmine/issues/9730


amousset:
Looks like the creation of the rudder user is missing in RPM packages.

peckpeck:
Commit modified

PR #1158, opened 2016-12-23
Fixes #9851: ncf-api-virtualenv build depends on selinux

peckpeck:
https://www.rudder-project.org/redmine/issues/9851


ncharles:
Based on @peckpeck this looks sensible

PR #1160, opened 2016-12-26
Fixes #9855: Add the performance patch to the CFEngine 3.10.0b1 to test actual policy execution performance

amousset:
https://www.rudder-project.org/redmine/issues/9855


ncharles:
<3

PR #1175, opened 2017-01-06
Fixes #9911: Wrong filename in apache conf packaging

amousset:
https://www.rudder-project.org/redmine/issues/9911


amousset:
Commit modified

PR #1185, opened 2017-01-10
Fixes #9941: Cannot install rudder-server-relay on CentOS

amousset:
https://www.rudder-project.org/redmine/issues/9941


rudder-server-relay/SPECS/rudder-server-relay.spec
```diff
@@ -76,6 +76,11 @@ BuildRoot: %{_tmppath}/%{name}-%{version}-%{release}-root-%(%{__id_u} -n)
 
 # Requirements
 
+# Disable dependency auto-generation, to prevent Python requirements
+# autodetection, which is not desired here.
+AutoReq: 0
+AutoProv: 0
```
amousset: Same as in `ncf-api-virtualenv`

PR #1189, opened 2017-01-11
Fixes #9956: Relay apache conf does not allow accessing the api

amousset:
https://www.rudder-project.org/redmine/issues/9956


rudder-server-relay/SOURCES/relay-api/apache/relay-api.conf
```diff
@@ -9,6 +9,11 @@ WSGISocketPrefix /var/run/wsgi
   Allow from all
 </Directory>
 
+<Directory /opt/rudder/share/relay-api/relay-api.wsgi>
```
peckpeck: File or use the parent name

amousset:
Commit modified

PR #1196, opened 2017-01-12
Fixes #9977: Wrong place for wsgi conf in deb packages

amousset:
https://www.rudder-project.org/redmine/issues/9977


amousset:
Commit modified

PR #1201, opened 2017-01-13
Fixes #9989: Relay api runs with www-data user instead of rudder

peckpeck:
https://www.rudder-project.org/redmine/issues/9989


peckpeck:
Commit modified

PR #1203, opened 2017-01-13
Fixes #9992: Use sudo to execute rudder remote run in the API

amousset:
https://www.rudder-project.org/redmine/issues/9992

### DIFF
--- a/rudder-server-relay/SOURCES/rudder-relay.sudo
+++ b/rudder-server-relay/SOURCES/rudder-relay.sudo
@@ -1,0 +1,2 @@
+# Allow the relay API to trigger remote runs
+rudder ALL = NOPASSWD: /opt/rudder/bin/rudder remote run *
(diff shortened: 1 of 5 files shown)


amousset:
Commit modified

rudder-server-relay/SOURCES/rudder-relay.sudo
```diff
@@ -0,0 +1,2 @@
+# Allow the relay API to trigger remote runs
+rudder ALL = NOPASSWD: /opt/rudder/bin/rudder remote run *
```
peckpeck: I think ins sudoers you just need to put a space to allow argument, not a star

amousset: From sudoers man page:

> If a Cmnd has associated command line arguments, then the arguments in the Cmnd must match exactly those given by the user on the command line (or match the wildcards if there are any).

PR #1204, opened 2017-01-13
Fixes #9993: ncf-api-venv user should not have access to a shell

amousset:
https://www.rudder-project.org/redmine/issues/9993


amousset:
Commit modified

PR #1212, opened 2017-01-19
Fixes #10035: Automatic install of jdk in the packaging fails on rhel

peckpeck:
https://www.rudder-project.org/redmine/issues/10035


amousset:
brilliant package naming...

PR #1214, opened 2017-01-20
Fixes #10040: Make webapp build on debian with java 8

peckpeck:
https://www.rudder-project.org/redmine/issues/10040


amousset:
tristitude

PR #1220, opened 2017-01-23
Fixes #10034: Error when upgrading rudder-agent

amousset:
https://www.rudder-project.org/redmine/issues/10034

### DIFF
--- a/rudder-agent/SOURCES/rudder-agent-postinst
+++ b/rudder-agent/SOURCES/rudder-agent-postinst
@@ -73,12 +73,15 @@ if [ "${RESTORE_SERVER_ROLES_BACKUP}" = "1" ]; then
   cp -r /var/backups/rudder/rudder-server-roles.conf ${CFE_DIR}/inputs/rudder-server-roles.conf
 fi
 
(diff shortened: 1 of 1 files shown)


rudder-agent/SOURCES/rudder-agent-postinst
```diff
@@ -73,12 +73,15 @@ if [ "${RESTORE_SERVER_ROLES_BACKUP}" = "1" ]; then
   cp -r /var/backups/rudder/rudder-server-roles.conf ${CFE_DIR}/inputs/rudder-server-roles.conf
 fi
 
+echo -n "INFO: Ensuring correct permissions on the keys directory..."
+# The chmod is a workaround for the limitations of rpm in rhel6 but it could be useful in other situations
+chmod 700 ${CFE_DIR}/ppkeys
```
ncharles: won't that fail if the path doesn't exist yet ?

ncharles: and why does the folder has its permission changed to 755 in the first place ?

amousset:
PR replaced by https://github.com/Normation/rudder-packages/pull/1222

PR #1231, opened 2017-01-30
Fixes #10090: Missing selinux-policy-devel in rudder-server-relay dependencies

amousset:
https://www.rudder-project.org/redmine/issues/10090


amousset:
OK, merging this PR

PR #1238, opened 2017-02-01
Fixes #10068: The user trying to open nodes list in relay-api is not rudder

amousset:
https://www.rudder-project.org/redmine/issues/10068


VinceMacBuche:
I'm trying the fix!!

VinceMacBuche:
@peckpeck Hey i was triying the fix before merge :( (comment above the label )

It works!

PR #1243, opened 2017-02-03
Fixes #10114: Allow relay-api to make asynchronous remote run call with output 

VinceMacBuche:
https://www.rudder-project.org/redmine/issues/10114


VinceMacBuche:
Commit modified

PR #1246, opened 2017-02-06
Fix #10144: Fusion inventory breaks on parsing last on wtmp line

armeniaca:


peckpeck:
Good, thank you.
Will merge it, but could you number your patch 24 instead of 19 ?
This is because it will be upmerged to branch 4.1 that contains other patches

armeniaca:
Hi @peckpeck ,
I renamed it to the issue number, so it will stay unique along all branches in future, I'd also suggest to continue with using the issue number in future ;-)

peckpeck:
You're completely right

PR #1251, opened 2017-02-09
Fixes #10162: rudder-relay has bad "sed" line

VinceMacBuche:
https://www.rudder-project.org/redmine/issues/10162


VinceMacBuche:
Commit modified

PR #1252, opened 2017-02-09
Fixes #10162: Rudder apache common conf is not packaged correctly on rpm

VinceMacBuche:
https://www.rudder-project.org/redmine/issues/10162


VinceMacBuche:
PR rebased

PR #1255, opened 2017-02-10
Fixes #10178: the shared-files directory is owned by root

peckpeck:
https://www.rudder-project.org/redmine/issues/10178


rudder-server-relay/SPECS/rudder-server-relay.spec
```diff
@@ -250,6 +250,9 @@ if ! getent passwd %{rudder_user} >/dev/null; then
   echo " Done"
 fi
 
+# change some directory to rudder owner
+chown rudder: /var/rudder/shared-files
```
amousset: Shouldn't we also `chmod 640` this directory ?

peckpeck:
Commit modified

PR #1259, opened 2017-02-15
Fixes #10073: Redirect http to https in relay api

amousset:
https://www.rudder-project.org/redmine/issues/10073

### DIFF
--- a/rudder-server-relay/SOURCES/rudder-apache-relay-nossl.conf
+++ b/rudder-server-relay/SOURCES/rudder-apache-relay-nossl.conf
@@ -1,0 +1,11 @@
+# Rudder webapp
+RewriteEngine on
+
+# If we are not currently connected via HTTP/S
+RewriteCond %{HTTPS} !=on
+
+# Restrict redirection to RudderRelay API 
+RewriteCond %{REQUEST_URI} ^/rudder/relay-api/?
+
+# Redirect to HTTPS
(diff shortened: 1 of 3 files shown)


peckpeck:
Shouldn't you remove equivalent redirection code in webap-nossl ?

Are all our call already ssl based ?

amousset:
we have some exceptions:

```
# Rule 2 - Don't use HTTP/S for these URLs to avoid breaking the compatibility for cURL
# clients (especially in Techniques)
RewriteCond %{REQUEST_URI} !^/rudder-(not-loaded|loading).html
RewriteCond %{REQUEST_URI} !^/uuid
RewriteCond %{REQUEST_URI} !^/inventories/?
RewriteCond %{REQUEST_URI} !^/inventory-updates/?
RewriteCond %{REQUEST_URI} !^/api/?
RewriteCond %{REQUEST_URI} !^/rudder/relay-api/?
```

All of this could go to common as we have a cond on http/https.

peckpeck:
is the webapp calling remote run directly via ssl ?

amousset:
yes: https://github.com/Normation/rudder/blob/master/rudder-web/src/main/resources/configuration.properties.sample#L75 and https://github.com/Normation/rudder-packages/blob/854a579624eabf76120a6c450e6e82f831c1805f/rudder-webapp/SOURCES/rudder-upgrade#L477

PR #1283, opened 2017-03-07
Fixes #10366: Add expiration on static resources

amousset:
https://www.rudder-project.org/redmine/issues/10366


fanf:
Regarding that PR and #1284, we discussed by oral about what brings https://www.rudder-project.org/redmine/issues/4519 (and PR #1568). It seems safer for now to let the webapp set the headers, and ounce we have some feedback, to make apache manage them or make other evolutions, so that we limit the moving parts these far in the release cycle. 

What do you thing?

amousset:
It is always better to set caching information in the application itself. The ncf-builder part is not covered by your change, but we can skip it for now.

PR #1286, opened 2017-03-08
Fixes #10375: Syntax error in parent ticket

amousset:
https://www.rudder-project.org/redmine/issues/10375


peckpeck:
lousy reviewer

PR #1287, opened 2017-03-08
Fixes #10373: Upgrading from 4.0 to 4.1 failed on Centos7.3, and purged LDAP directory

ncharles:
https://www.rudder-project.org/redmine/issues/10373

### DIFF
--- a/rudder-inventory-ldap/SOURCES/rudder-slapd-configure
+++ b/rudder-inventory-ldap/SOURCES/rudder-slapd-configure
@@ -1,0 +1,63 @@
+#!/bin/sh
+
+# Configure maxsize in /opt/rudder/etc/openldap/slapd.conf based on
+# content of /etc/default/rudder-slapd, or auto-computed based on available memory
+
+# Script specific
+PROG_NAME='rudder-slapd'
+OS=`uname -s`   # To adapt message printing
+
+
+#====================================================================
+# Message function
+#====================================================================
+message() {
+  # $1: syslog level
+  # $2: message
+
+  if [ $OS = "Linux" ]
+  then
+    logger -p "local4.$1" -s -t $PROG_NAME -i "$2"
+  else
+    # Try without option -s
+    logger -p "local4.$1" -t $PROG_NAME -i "$2"
+    echo "$PROG_NAME: $2"
+  fi
+}
+
+#====================================================================
+
+
+
+#====================================================================
+# Load specific parameters
+#====================================================================
+. /opt/rudder/etc/rudder-slapd.conf
+
+if [ -f /etc/default/$PROG_NAME ]
+then
+  . /etc/default/$PROG_NAME
+  message "info" "[INFO] Using /etc/default/$PROG_NAME for configuration"
+fi
+
+# configuration of mdbsize
+if [ "${RUDDER_MDBSIZE}" != "noauto" ]
+then
+  if [ "${RUDDER_MDBSIZE}" = "auto" ] || [ "${RUDDER_MDBSIZE}" = "" ]
+  then
+    # calculate the value to match about 50% of RAM (this is only virtual memory)
+    MEMSIZE=$(sed -n '/^MemTotal/s/[^0-9]//gp' /proc/meminfo)
+    MDBSIZE=$((${MEMSIZE} * 1024 / 2 ))
+    # on 32 bits architectures max it to 500M (PAE allows more ram but not more virtual memory)
+    if [ "$(getconf LONG_BIT)" -ne "64" ] && [ "${MDBSIZE}" -gt 524288000 ]
+    then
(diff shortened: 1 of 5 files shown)


rudder-inventory-ldap/debian/postinst
```diff
@@ -77,26 +77,33 @@ case "$1" in
...
-				# Start OpenLDAP
-				echo -n "INFO: Starting rudder-slapd..."
-				invoke-rc.d rudder-slapd start >/dev/null 2>&1
-				echo " Done"
-
-				echo "INFO: OpenLDAP database was successfully upgraded to new format"
-
-				if [ -x /opt/rudder/bin/rudder-upgrade-ldap ]
+				if [ $? -ne 0 ]
 				then
```
peckpeck: This script is set -e, you will never reach this step

peckpeck: put the command directly in the if instead

ncharles:
Commit modified

ncharles:
Commit modified

rudder-inventory-ldap/SOURCES/rudder-slapd-configure
```diff
@@ -0,0 +1,63 @@
...
+  if [ "${RUDDER_MDBSIZE}" = "auto" ] || [ "${RUDDER_MDBSIZE}" = "" ]
+  then
+    # calculate the value to match about 50% of RAM (this is only virtual memory)
+    MEMSIZE=$(sed -n '/^MemTotal/s/[^0-9]//gp' /proc/meminfo)
+    MDBSIZE=$((${MEMSIZE} * 1024 / 2 ))
+    # on 32 bits architectures max it to 500M (PAE allows more ram but not more virtual memory)
+    if [ "$(getconf LONG_BIT)" -ne "64" ] && [ "${MDBSIZE}" -gt 524288000 ]
+    then
+      MDBSIZE=524288000
+    fi
```
jooooooon: This exception for 32 bit machines is no longer applicable - MDBSIZE is not a measure of RAM.

rudder-inventory-ldap/SOURCES/rudder-slapd-configure
```diff
@@ -0,0 +1,63 @@
...
+fi
+
+# configuration of mdbsize
+if [ "${RUDDER_MDBSIZE}" != "noauto" ]
+then
+  if [ "${RUDDER_MDBSIZE}" = "auto" ] || [ "${RUDDER_MDBSIZE}" = "" ]
+  then
+    # calculate the value to match about 50% of RAM (this is only virtual memory)
+    MEMSIZE=$(sed -n '/^MemTotal/s/[^0-9]//gp' /proc/meminfo)
+    MDBSIZE=$((${MEMSIZE} * 1024 / 2 ))
```
jooooooon: MDBSIZE is the total size of the underlying database - it should not depend on RAM, but on the number of nodes or a similar metric. Or just be huge so that there is "no limit".

peckpeck: The database is mapped into virtual memory, thus cannot physically exceed 4G on 32bit and in practice, the value is more likely to be 2G

ncharles: and the actual database must fit in RAM (maxsize is the upper bound of the database)

ncharles: ha, dismiss this comment, it's wrong (some docs says it's related to memory usage, but actuall doc from symas https://symas.com/understanding-lmdb-database-file-sizes-and-memory-utilization/ "LMDB uses memory-mapped files to make the entire database appear as if it’s in primary memory, even if the database itself is larger than the total complement of RAM on the system."

rudder-inventory-ldap/SPECS/rudder-inventory-ldap.spec
```diff
@@ -292,25 +296,31 @@ if [ -n "${BACKUP_LDIF}" ]; then
...
-
-		echo "INFO: OpenLDAP database was successfully upgraded to new format"
-
-		if [ -x /opt/rudder/bin/rudder-upgrade-ldap ]
+		# Import the backed up database
+		if /opt/rudder/sbin/slapadd -q -l ${BACKUP_LDIF}
 		then
-			echo "INFO: Running the Rudder upgrade script to replay LDAP migrations on the old database content..."
-			/opt/rudder/bin/rudder-upgrade-ldap
+			echo "ERROR: Failed to restore data from old format into the new format"
```
ncharles: this is exactly the opposite: if command then success else failure fi

peckpeck: indeed, or if ! command

PR #1291, opened 2017-03-14
Fixes #10417: sharedfiles are never cleaned

ncharles:
https://www.rudder-project.org/redmine/issues/10417


ncharles:
Commit modified

PR #1293, opened 2017-03-15
Fixes #10424: LDAP error at upgrade - Cannot allocate memory

amousset:
https://www.rudder-project.org/redmine/issues/10424

### DIFF
--- a/rudder-inventory-ldap/SOURCES/rudder-inventory-ldap.init
+++ b/rudder-inventory-ldap/SOURCES/rudder-inventory-ldap.init
@@ -174,6 +174,7 @@ BACKUP_SUFFIX="`date +%Y%m%d%H%M%S`.ldif"
 # Other
 TIMEOUT="60"    # Max time to stop process
 FD_LIMIT="2048" # Max file descriptor
+VM_LIMIT="unlimited" # Virtual memory limit
 
 # Script specific
 PROG_NAME=`basename $0 | sed 's/^[KS][0-9][0-9]//'` # For nice messages
@@ -350,6 +351,21 @@ if [ $PORT -lt 1024 -a $MYUID -ne 0 ]
 then
   message "alert" "[ALERT] only root can launch OpenLDAP on port $PORT"
   exit 1
+fi
+
+# Virtual memory limit, only for root
+# Needed to manipulate big memory-mapped files
+if [ $MYUID -eq 0 ]
+then
+  ulimit -v $VM_LIMIT
+  if [ $? -eq 0 ]
+  then
+    message "info" "[OK] virtual memory limit set to $VM_LIMIT"
+  else
+    message "warning" "[WARNING] Fail to set file virtual memory to $VM_LIMIT, going to next step"
+  fi
+else
+  message "info" "[INFO] memory limit not modified (require root privileges)"
 fi
 
 #====================================================================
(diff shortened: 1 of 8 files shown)


rudder-inventory-ldap/SOURCES/rudder-inventory-ldap.init
```diff
@@ -174,6 +174,7 @@ BACKUP_SUFFIX="`date +%Y%m%d%H%M%S`.ldif"
 # Other
 TIMEOUT="60"    # Max time to stop process
 FD_LIMIT="2048" # Max file descriptor
+VM_LIMIT="108490720" # Max virtual memory
```
amousset: Not sure about what to put here.

amousset:
Commit modified

rudder-inventory-ldap/SOURCES/rudder-inventory-ldap.init
peckpeck: you should put the mdb size plus something i think

peckpeck: good idea

amousset: We can't easily read the mdb size from the config file as it is written during service start

peckpeck: Yes, but if i understand correctly, this value is the maximum virtual memory allowed to be associated  to the process, which includes mapped files.

So this has to be dependent or far above mdb size (ie unlimited).
So you could compute it again, or simply put unlimited here

amousset: I will move this setting to rudder-slapd.conf, allowing override from the default file

amousset: We could make the ulimit change a part of the script that configures the mdb size, but it not called from every action in the init script (and that would not make sense to add it).

We can:

* Put a `ulimit -v` before every slap* call, it could be
  * a hardcoded value
  * computed from a script that uses the current mdbsize, which would be called in migration scripts and in every action fo the init script
* The only other option I see is to configure virtual memory limit for the root user in limits.conf. I don't think we want to to it silently, or even do it at all.

peckpeck: This is the init script, every thing is done here.
So you could just put ulimit at the beginning of the script and forget everything else. 
And it is probably necessary to put the same command in the postinstall script that calls slapcat et al.

amousset: Yes but at the beginning of the script the mdb size may not be computed already.

The simplest solution:

* `ulimit -v unlimited` at the beginning of the init script
* `ulimit -v unlimited` at the beginning of the migration script in postinstall

Virtual memory seems unlimited by default on modern OSes.

amousset:
Commit modified

rudder-inventory-ldap/SOURCES/rudder-inventory-ldap.init
```diff
@@ -174,6 +174,7 @@ BACKUP_SUFFIX="`date +%Y%m%d%H%M%S`.ldif"
 # Other
 TIMEOUT="60"    # Max time to stop process
 FD_LIMIT="2048" # Max file descriptor
+$VM_LIMIT="unlimited" # Virtual memory limit
```
peckpeck: $ ?

amousset:
Commit modified

PR #1306, opened 2017-03-21
Fixes #10427: On freshly installed centos7, ldap connections fail with "bad auth"

VinceMacBuche:
https://www.rudder-project.org/redmine/issues/10427

### DIFF
--- a/rudder-inventory-ldap/SPECS/rudder-inventory-ldap.spec
+++ b/rudder-inventory-ldap/SPECS/rudder-inventory-ldap.spec
@@ -287,7 +287,12 @@ if [ -n "${BACKUP_LDIF}" ]; then
 		# Stop OpenLDAP - use forcestop to avoid the init script failing
 		# when trying to do the backup with bad libdb versions
 		echo -n "INFO: Stopping rudder-slapd..."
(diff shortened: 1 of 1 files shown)


rudder-inventory-ldap/SPECS/rudder-inventory-ldap.spec
```diff
@@ -287,7 +287,12 @@ if [ -n "${BACKUP_LDIF}" ]; then
 		# Stop OpenLDAP - use forcestop to avoid the init script failing
 		# when trying to do the backup with bad libdb versions
 		echo -n "INFO: Stopping rudder-slapd..."
-		service rudder-slapd forcestop >/dev/null 2>&1
+      %if 0%{?rhel} < 7
```
peckpeck: what about non rhel distro, like sles ?

VinceMacBuche: Hmm, we have the same line in lot of ther packages (reports / relay ... ) I think it's using the first one

VinceMacBuche: Since it is not define, it will be 0 < 7

PR #1307, opened 2017-03-21
Fixes #10426: Apache not started on a fresh centos7 install (selinux problem)

VinceMacBuche:
https://www.rudder-project.org/redmine/issues/10426


amousset:
This should have been done in 3.1.

PR #1313, opened 2017-03-23
Fixes #10491: selinux policy should be applied at the end of postinstall or techniqe  folder will not be managed correctly

VinceMacBuche:
https://www.rudder-project.org/redmine/issues/10491


VinceMacBuche:
Commit modified

VinceMacBuche:
OK, merging this PR

PR #1316, opened 2017-03-24
Fixes #10475: Empty output of service rudder restart on Debian 8

peckpeck:
https://www.rudder-project.org/redmine/issues/10475


peckpeck:
This is not the problem and this line has already been fixed

PR #1319, opened 2017-03-27
Fixes #10506: Rudder 4.1 fails to install Ubuntu/Debian because of rudder-slapd service restart

peckpeck:
https://www.rudder-project.org/redmine/issues/10506

### DIFF
--- a/rudder-inventory-ldap/debian/postinst
+++ b/rudder-inventory-ldap/debian/postinst
@@ -65,6 +65,10 @@ case "$1" in
 					exit 1
 				fi
 
+        # Reload systemd when it is available so that it creates sysv-init shims
+        # This is necessary to make the service command work, and will be obsolete when we create systemd units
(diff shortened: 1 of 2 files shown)


rudder-inventory-ldap/debian/postinst
```diff
@@ -65,6 +65,10 @@ case "$1" in
 					exit 1
 				fi
 
+        # Reload systemd when it is available so that it creates sysv-init shims
+        # This is necessary to make the service command work, and will be obsolete when we create systemd units
+        [ -x /bin/systemctl ] && /bin/systemctl daemon-reload
```
ncharles: this should also be of the if block - error at install is between INFO: Restarting syslogd... and INFO: Restarting rudder-slapd... , so not in this block. It's probably safer to add it on line 40

peckpeck: It is not necessary for syslog since it has its own unit file.

PR #1323, opened 2017-03-28
Fixes #10521: Error in migration (ldap backup not found because ending by .gz)

peckpeck:
https://www.rudder-project.org/redmine/issues/10521


peckpeck:
OK, merging this PR

PR #1328, opened 2017-04-10
Fixes #10559: On upgrade from 3.1 to 4.1, policy could not be generated any more

ncharles:
https://www.rudder-project.org/redmine/issues/10559


ncharles:
OK, merging this PR

ncharles:
OK, merging this PR

PR #1334, opened 2017-04-12
Fixes #10595: Broken tests after parent

amousset:
https://www.rudder-project.org/redmine/issues/10595


amousset:
OK, merging this PR

PR #1349, opened 2017-05-23
Fixes #10776: Share promises for windows agents in https

peckpeck:
https://www.rudder-project.org/redmine/issues/10776


amousset:
I wonder if the apache conf should be distributed in a plugin instead.

peckpeck:
PR replaced by https://github.com/Normation/rudder-packages/pull/1350

PR #1353, opened 2017-06-06
Fixes #10771: check-rudder-agent doesn't purge the lmdb file when run via cron

peckpeck:
https://www.rudder-project.org/redmine/issues/10771


peckpeck:
OK, merging this PR

PR #1357, opened 2017-06-09
Fixes #10885: /var/rudder/share must be readable by rudder-policy-reader

peckpeck:
https://www.rudder-project.org/redmine/issues/10885


peckpeck:
Commit modified

PR #1361, opened 2017-06-22
Fixes #10939: Rudder Version in rudder-pkg is hardcoded

fanf:
https://www.rudder-project.org/redmine/issues/10939


fanf:
Not the correct correction

PR #1367, opened 2017-07-03
Fixes #10990: When installing Rudder-Webapp, postinst script runs rudder-upgrade before initializing the git, and it can fail

amousset:
https://www.rudder-project.org/redmine/issues/10990

### DIFF
--- a/rudder-webapp/debian/postinst
+++ b/rudder-webapp/debian/postinst
@@ -46,13 +46,6 @@ case "$1" in
   service apache2 restart >/dev/null 2>&1
   echo " Done"
 
-  # Run any upgrades
-  # Note this must happen *before* creating the technique store, as it was moved in version 2.3.2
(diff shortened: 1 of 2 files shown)


rudder-webapp/debian/postinst
```diff
@@ -88,6 +81,13 @@ case "$1" in
     fi
   fi
 
+  # Run any upgrades
+  # Note this must happen *before* creating the technique store, as it was moved in version 2.3.2
```
ncharles: you could remove this comment, can't you ?

amousset:
Commit modified

rudder-webapp/debian/postinst
amousset: Comments that say the contrary of the code are always a pleasure, we should keep them 🤦‍

PR #1368, opened 2017-07-03
Fixes #11064: if plugin is broken, we cannot upgrade it

amousset:
https://www.rudder-project.org/redmine/issues/11064


amousset:
OK, merging this PR

PR #1369, opened 2017-07-05
Fixes #11045: When upgrading 4.2 server, system techniques from plugin are purged

amousset:
https://www.rudder-project.org/redmine/issues/11045


ncharles:
image

PR #1370, opened 2017-07-24
Fixes #11161: Share file from shared folder using relay rest api

VinceMacBuche:
https://www.rudder-project.org/redmine/issues/11161

### DIFF
--- a/rudder-server-relay/SOURCES/relay-api/relay_api/views.py
+++ b/rudder-server-relay/SOURCES/relay-api/relay_api/views.py
@@ -1,5 +1,5 @@
 from relay_api import app
-from relay_api.shared_files import shared_files_put, shared_files_head, shared_files_put_forward, shared_files_head_forward
+from relay_api.shared_files import shared_files_put, shared_files_head, shared_files_put_forward, shared_files_head_forward, shared_folder_head
 from relay_api.remote_run import remote_run_generic
 from relay_api.common import *
 
@@ -44,6 +44,20 @@ def put_file(target_uuid, source_uuid, file_id):
   except Exception as e:
     return format_error(e, API_DEBUGINFO)
 
+
+@app.route("/shared-folder/<path:file_name>", methods=["HEAD"])
+def head_shared_folder(file_name):
+  try:
+    hash_type = request.args["hash_type"]
+    if hash_type is None:
+      hash_type = "sha256"
+    file_hash = request.args["hash"]
+    return_code = shared_folder_head(file_name, file_hash, hash_type)
+    return format_response("", return_code)
+
+  except Exception as e:
+    print(traceback.format_exc())
+    return format_error(e, API_DEBUGINFO)
 
 @app.route('/shared-files/<string:target_uuid>/<string:source_uuid>/<string:file_id>', methods=['HEAD'])
 def head_file(target_uuid, source_uuid, file_id):
--- a/rudder-server-relay/SOURCES/rudder-apache-relay-ssl.conf
+++ b/rudder-server-relay/SOURCES/rudder-apache-relay-ssl.conf
@@ -34,3 +34,22 @@ RewriteRule /policies/(.*?)/(.*) /var/rudder/share/$1/$2
     Include /opt/rudder/etc/rudder-networks-24.conf
   </IfVersion>  
 </Directory>
+
+
+<Location /rudder/relay-api/shared-folder>
+  SSLVerifyClient require
+  SSLUserName SSL_CLIENT_S_DN_CN
+  SSLRequireSSL
+
+  <IfVersion < 2.4>
+    Order allow,deny
+    Include /opt/rudder/etc/rudder-networks.conf
+  </IfVersion>
+  <IfVersion >= 2.4>
+    Include /opt/rudder/etc/rudder-networks-24.conf
+  </IfVersion>
+</Location>
+
+RewriteCond %{REQUEST_METHOD}  =GET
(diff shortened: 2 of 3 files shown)


rudder-server-relay/SOURCES/relay-api/relay_api/views.py
```diff
@@ -1,12 +1,12 @@
...
-from relay_api.shared_files import shared_files_put, shared_files_head, shared_files_put_forward, shared_files_head_forward
+from relay_api.shared_files import shared_files_put, shared_files_head, shared_files_put_forward, shared_files_head_forward, shared_folder_head
 from relay_api.remote_run import remote_run_generic
 from relay_api.common import *
 
 from flask import Flask, jsonify, request, abort, make_response
 from StringIO import StringIO
 import traceback
-
+import os
```
peckpeck: what for ?

rudder-server-relay/SOURCES/rudder-apache-relay-ssl.conf
```diff
@@ -34,3 +34,35 @@ RewriteRule /policies/(.*?)/(.*) /var/rudder/share/$1/$2
...
+
+</Location>
+
+RewriteCond %{REQUEST_METHOD}  =GET
+RewriteRule /rudder/relay-api/shared-folder/(.*) /var/rudder/configuration-repository/shared-files/$1
+
+<Directory /var/rudder/configuration-repository/shared-files>
+  SSLVerifyClient require
+  SSLUserName SSL_CLIENT_S_DN_CN
+  SSLRequireSSL
```
peckpeck: This should go into location, otherwise the HEAD will not be authenticated

rudder-server-relay/SOURCES/rudder-apache-relay-ssl.conf
```diff
@@ -34,3 +34,35 @@ RewriteRule /policies/(.*?)/(.*) /var/rudder/share/$1/$2
...
+  <IfVersion >= 2.4>
+    Include /opt/rudder/etc/rudder-networks-24.conf
+  </IfVersion>
+
+</Location>
+
+RewriteCond %{REQUEST_METHOD}  =GET
+RewriteRule /rudder/relay-api/shared-folder/(.*) /var/rudder/configuration-repository/shared-files/$1
+
+<Directory /var/rudder/configuration-repository/shared-files>
```
peckpeck: This is not necessary since the location says the same thing

rudder-server-relay/SOURCES/relay-api/relay_api/views.py
VinceMacBuche: relicate when testing ! to remove

rudder-server-relay/SOURCES/rudder-apache-relay-ssl.conf
VinceMacBuche: The location does not say the same thing, We need to define access to the directory through apache

rudder-server-relay/SOURCES/rudder-apache-relay-ssl.conf
VinceMacBuche: You're right this should go to location too

VinceMacBuche:
Commit modified

rudder-server-relay/SOURCES/rudder-apache-relay-ssl.conf
peckpeck: No, both are checked by apache.
It checks Location when hitting this location. Then, aliases and rewrites are done, then if is matches a path on the server, it checks Directory and then it check File.
So in this case it will hit both location and Directory which is redundant

VinceMacBuche:
Commit modified

VinceMacBuche:
Commit modified

matya:
Why are you using md5? I mean, the whole world is deprecating md5...

VinceMacBuche:
Commit modified

matya:
Thank you in the name of all current and future users!

rudder-server-relay/SOURCES/relay-api/relay_api/views.py
```diff
@@ -45,6 +45,19 @@ def put_file(target_uuid, source_uuid, file_id):
     return format_error(e, API_DEBUGINFO)
 
 
+@app.route("/shared-folder/<path:file_name>", methods=["HEAD"])
+def head_shared_folder(file_name):
+  try:
+    file_hash = request.args["hash"]
```
peckpeck: Tjhere should be a hash type parameter

VinceMacBuche:
Commit modified

rudder-server-relay/SOURCES/relay-api/relay_api/views.py
```diff
@@ -45,6 +45,22 @@ def put_file(target_uuid, source_uuid, file_id):
...
 
+@app.route("/shared-folder/<path:file_name>", methods=["HEAD"])
+def head_shared_folder(file_name):
+  try:
+    hash_type = request.args["hash_type"]
+    if hash_type is None:
+      hash_type = "sha256"
+    file_hash = request.args["hash"]
+    if file_hash is None:
+      file_hash = ""
```
peckpeck: don't we already know that it will fail ?

VinceMacBuche: It will fail, but i choose to delegate all error checking in the method from shared-files and stay as simple as possible.

We can fail directly if you prefer

VinceMacBuche:
Commit modified

VinceMacBuche:
Commit modified

PR #1373, opened 2017-08-03
Fixes #10963: Patch for "Unable to setup metadata virtual space list" not in the correct place

amousset:
https://www.rudder-project.org/redmine/issues/10963

### DIFF
--- a/rudder-jetty/SOURCES/Makefile
+++ b/rudder-jetty/SOURCES/Makefile
@@ -70,6 +70,7 @@ localdepends: ./jetty7
 
 	cp ./jetty7/bin/jetty-rpm.sh ./jetty7/bin/jetty-sles.sh
 	patch -p0 -s < patches/jetty/jetty-init-sles.patch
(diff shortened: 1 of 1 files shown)


amousset:
Commit modified

rudder-jetty/SOURCES/Makefile
```diff
@@ -70,6 +70,7 @@ localdepends: ./jetty7
 
 	cp ./jetty7/bin/jetty-rpm.sh ./jetty7/bin/jetty-sles.sh
 	patch -p0 -s < patches/jetty/jetty-init-sles.patch
+	patch -p0 -s < patches/jetty/jetty-init-ulimit.patch
```
peckpeck: In the issue: "and EVEN with the patch, the parent bug is not corrected." how is this addressed ?

amousset: The ticket says it was patched in /opt/rudder./... whcih is the source of what is copied to /etc/init.d/rudder-jetty, so I think this patch will be enough actually.

PR #1384, opened 2017-09-14
Fixes #11358: ERROR 405 when using "File from shared folder" generic method

VinceMacBuche:
https://www.rudder-project.org/redmine/issues/11358


ncharles:
shouldn't it go to 4.1 as well ?

VinceMacBuche:
Commit modified

VinceMacBuche:
Not in 4.1, shared-folder api only exists in 4.2. However we should check shared-file api and fix it in 4.1 if needed

PR #1391, opened 2017-10-04
Fixes #11125: When we have a /var full, Rudder LDAP or Postgres can be corrupted

peckpeck:
https://www.rudder-project.org/redmine/issues/11125


amousset:
I think this needs to be a lot more visible to the user. At least in some log files, or even better in the html error message that will be displayed.

peckpeck:
Commit modified

peckpeck:
OK, merging this PR

PR #1394, opened 2017-10-19
Fixes #11626: Build on Deb7 and ubuntu14.04 fails because jdk8u144 is unavalaible

VinceMacBuche:
https://www.rudder-project.org/redmine/issues/11626


VinceMacBuche:
Commit modified

PR #1396, opened 2017-10-19
Fixes #11628: Error at installation of Rudder 4.1 on centos 6

amousset:
https://www.rudder-project.org/redmine/issues/11628


amousset:
Commit modified

PR #1397, opened 2017-10-20
Fixes #11647: Upgrade to Jetty9

fanf:
https://www.rudder-project.org/redmine/issues/11647

Nice novelty: in Jetty9, we have a $JETTY_BASE where you define your modules (new concept, very cleanly defined!) and webapp env config. $JETTY_HOME is kept only for the base distrib and is not modifed (which should ease future upgrade). Doc is here http://www.eclipse.org/jetty/documentation/current/startup-base-and-home.html, and it is nice! The one for module is here, about how to list/search/add modules: http://www.eclipse.org/jetty/documentation/current/startup-modules.html.

All patches pass with minor changes. I'm not sure we need to keep the one about setuid, because there is a module which seem to do just that. 

I choose to define JETTY_BASE=/opt/rudder/etc/rudder-jetty-base (so along /opt/rudder/etc/rudder-jetty.config file). /opt/rudder/share/webapp does not change.


rudder-jetty/SPECS/rudder-jetty.spec
```diff
@@ -141,12 +141,12 @@ mkdir -p %{buildroot}/var/rudder/run
 
 cd %{_topdir}/SOURCES
 
-cp -a jetty7 %{buildroot}/opt/rudder
+cp -a jetty %{buildroot}/opt/rudder
```
peckpeck: You should also copy jetty-base files here

fanf:
PR rebased

PR #1403, opened 2017-11-08
Fixes #11712: Change rudder-techniques ncf depdency to a correct version

VinceMacBuche:
https://www.rudder-project.org/redmine/issues/11712


peckpeck:
I think this should not be upmerged.
So I will let you merge this manually

PR #1406, opened 2017-11-15
Fixes #11727: Rudder installation fails on SLES12

peckpeck:
https://www.rudder-project.org/redmine/issues/11727

### DIFF
--- a/rudder-agent/SPECS/rudder-agent.spec
+++ b/rudder-agent/SPECS/rudder-agent.spec
@@ -350,6 +350,10 @@ then
 %endif
   CFRUDDER_FIRST_INSTALL=1
 fi
+# mandatory with systemd wrapper for old init
+%if 0%{?suse_version} && 0%{?suse_version} >= 1315
+systemctl daemon-reload
+%endif
 
 %if "%{?rhel}" == "3"
 # Update and reload ld.so configuration if needed, RHEL/CentOS 3 style.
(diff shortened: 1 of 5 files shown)


rudder-agent/SPECS/rudder-agent.spec
```diff
@@ -350,6 +350,10 @@ then
 %endif
   CFRUDDER_FIRST_INSTALL=1
 fi
+# mandatory with systemd wrapper for old init
+%if 0%{?suse_version} && 0%{?suse_version} >= 1315
```
peckpeck: This is sles12 and later, @amousset dou you think the condition should be larger ?

ncharles: isn't it something we should do on other distro as well (debian9, rhel 7) ?

peckpeck: It it tied to how chkconfig --add and update-rc.d is done by the distro (this command should reload by itself). So this is distro specific and it has not been reported on other distro yet.
That's why i'm asking for comment from @amousset

rudder-reports/SPECS/rudder-reports.spec
```diff
@@ -181,6 +181,9 @@ chkconfig --add ${POSTGRESQL_SERVICE_NAME} >/dev/null 2>&1
   chkconfig ${POSTGRESQL_SERVICE_NAME} on >/dev/null 2>&1
 %endif
 echo " Done"
+# mandatory with systemd wrapper for old init
+%if 0%{?suse_version} && 0%{?suse_version} >= 1315
+systemctl daemon-reload
```
amousset: missing `%endif`

peckpeck:
Commit modified

PR #1413, opened 2017-11-17
Fixes #11747: When we have a /var full, Rudder LDAP or Postgres can be corrupted

peckpeck:
https://www.rudder-project.org/redmine/issues/11747


peckpeck:
OK, merging this PR

PR #1414, opened 2017-11-20
Fixes #11752: Add a curl dependency in the agent

amousset:
https://www.rudder-project.org/redmine/issues/11752


amousset:
Commit modified

PR #1424, opened 2017-12-11
Fixes #10454: Rudder root on SLES12 shows LDAP error during installation

amousset:
https://www.rudder-project.org/redmine/issues/10454


amousset:
OK, merging this PR

PR #1435, opened 2018-01-09
Fixes #11935: Broken build on rhel

peckpeck:
https://www.rudder-project.org/redmine/issues/11935

### DIFF
--- a/rudder-agent/SPECS/rudder-agent.spec
+++ b/rudder-agent/SPECS/rudder-agent.spec
@@ -131,6 +131,14 @@ Requires: kernel-utils
 Requires: dmidecode
 %endif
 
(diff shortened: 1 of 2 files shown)


peckpeck:
Commit modified

rudder-agent/SPECS/rudder-agent.spec
```diff
@@ -131,6 +131,14 @@ Requires: kernel-utils
 Requires: dmidecode
 %endif
 
+# https fails on old distro because they don't support modern certificates (namely rhel3 and sles10)
```
amousset: don't we need to the the default value to true?

peckpeck: it can be anything, the default within the makefile is true since we compare to the litteral "false"

PR #1444, opened 2018-01-15
Fixes #11965: Broken server install on CentOS7 in 4.3

amousset:
https://www.rudder-project.org/redmine/issues/11965


amousset:
Commit modified

PR #1447, opened 2018-01-16
Fixes #11972: pre-inst scripts should be part of the package

amousset:
https://www.rudder-project.org/redmine/issues/11972


amousset:
PR replaced by https://github.com/Normation/rudder-packages/pull/1448

PR #1460, opened 2018-01-19
Fixes #12017: Perl system dependencies are not installed soon enough

peckpeck:
https://www.rudder-project.org/redmine/issues/12017


peckpeck:
Commit modified

PR #1465, opened 2018-01-22
Stop jetty7 service in preinst to prevent it from staying after upgrade

amousset:
https://www.rudder-project.org/redmine/issues/12000


amousset:
Commit modified

amousset:
Commit modified

PR #1476, opened 2018-01-24
Fixes #12040: Inventory fails to run (at least on centos7)

peckpeck:
https://www.rudder-project.org/redmine/issues/12040

### DIFF
--- a/rudder-agent/SOURCES/Makefile
+++ b/rudder-agent/SOURCES/Makefile
@@ -487,9 +487,9 @@ CPANM:=$(CURDIR)/perl-$(PERL_VERSION)/App-cpanminus/bin/cpanm
 MODULE_INSTALL:=MAKEFLAGS= $(PERL) $(CPANM) $(CPAN_OPT) -L $(INSTALL_DIR) --from $(CPAN_URL) --notest
 # List from README / Makefile.PL in fusion-inventory-agent
 # -> requires + inventory and network recommends
-FUSION_MODULES = File::Which LWP::UserAgent Net::IP Text::Template UNIVERSAL::require XML::TreePP Parse::EDID DateTime DateTime::TimeZone DateTime::TimeZone::Local::Unix Net::SNMP Net::NBName Thread::Queue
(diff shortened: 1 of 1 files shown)


rudder-agent/SPECS/rudder-agent.spec
```diff
@@ -113,28 +113,38 @@ Conflicts: rudder-agent
...
+Requires: perl-IPC-Cmd perl-File-Which net-snmp-perl
+BuildRequires: perl-IPC-Cmd perl-File-Which net-snmp-perl
 %endif
 
-# rhel perl core is too minimal, we try to not add too much here
+# rhel 7 perl core is too minimal
 %if 0%{?rhel} && 0%{?rhel} >= 7
-Requires: perl-Digest 
-BuildRequires: perl-Digest
+Requires: perl-Digest perl-DateTime-TimeZone perl-Thread-Queue perl-Data-Dumper
```
ncharles: didn't we said that we were to left the dependencies as they are, and see when it would fail ?

rudder-agent/SOURCES/Makefile
```diff
@@ -487,9 +487,9 @@ CPANM:=$(CURDIR)/perl-$(PERL_VERSION)/App-cpanminus/bin/cpanm
 MODULE_INSTALL:=MAKEFLAGS= $(PERL) $(CPANM) $(CPAN_OPT) -L $(INSTALL_DIR) --from $(CPAN_URL) --notest
 # List from README / Makefile.PL in fusion-inventory-agent
 # -> requires + inventory and network recommends
-FUSION_MODULES = File::Which LWP::UserAgent Net::IP Text::Template UNIVERSAL::require XML::TreePP Parse::EDID DateTime DateTime::TimeZone DateTime::TimeZone::Local::Unix Net::SNMP Net::NBName Thread::Queue
+FUSION_MODULES = File::Which LWP::UserAgent Net::IP Text::Template UNIVERSAL::require XML::TreePP Parse::EDID DateTime DateTime::TimeZone DateTime::TimeZone::Local::Unix Net::SNMP Net::NBName Thread::Queue Data::Dumper
```
ncharles: this did not work
Dumper is now in /opt/rudder/lib/perl5/Module/Build/Dumper.pm, but it is expected in /opt/rudder/lib/perl5/Data/

PR #1477, opened 2018-01-25
Fixes #12051: Unpackaged post inst script on ubuntu 16.04

amousset:
https://www.rudder-project.org/redmine/issues/12051

### DIFF
--- a/rudder-server-root/debian/install
+++ b/rudder-server-root/debian/install
@@ -1,0 +1,3 @@
+opt/rudder
+etc
+var/rudder


rudder-server-root/debian/dirs
```diff
@@ -1,3 +1,3 @@
 opt/rudder/etc
 opt/rudder/bin
-
+opt/rudder/share
```
ncharles: this does not exist on the agent side, so I doubt the it will solve the issue

amousset:
Commit modified

amousset:
Commit modified

PR #1499, opened 2018-02-09
Fixes #11153: sysconfig/apache2 modifications get removed

peckpeck:
https://www.rudder-project.org/redmine/issues/11153

### DIFF
--- a/rudder-server-relay/SPECS/rudder-server-relay.spec
+++ b/rudder-server-relay/SPECS/rudder-server-relay.spec
@@ -318,9 +318,16 @@ fi
 
 %if 0%{?suse_version}
 # Add required includes in the apache2 configuration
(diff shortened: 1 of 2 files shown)


rudder-server-relay/SPECS/rudder-server-relay.spec
```diff
@@ -320,6 +320,7 @@ fi
 # Add required includes in the apache2 configuration
 if ! grep -qE "^. /etc/sysconfig/rudder-relay-apache$" /etc/sysconfig/apache2; then
```
amousset: this will not fix existing installs, shouldn't be a separate check for the variable?

peckpeck:
Commit modified

peckpeck:
Replaced by https://github.com/Normation/rudder-packages/pull/1509

PR #1506, opened 2018-02-15
Fixes #12117: rudder-inventory-ldap can't be installed on centos (7 at least)

peckpeck:
https://www.rudder-project.org/redmine/issues/12117


peckpeck:
Commit modified

PR #1508, opened 2018-02-15
Fixes #12121: Wrong condition to check if PID directory exists when stopping ldap

VinceMacBuche:
https://www.rudder-project.org/redmine/issues/12121


VinceMacBuche:
Commit modified

PR #1509, opened 2018-02-16
Fixes #11153: sysconfig/apache2 modifications get removed

amousset:
https://www.rudder-project.org/redmine/issues/11153

### DIFF
--- a/rudder-server-relay/SPECS/rudder-server-relay.spec
+++ b/rudder-server-relay/SPECS/rudder-server-relay.spec
@@ -318,9 +318,16 @@ fi
 
 %if 0%{?suse_version}
 # Add required includes in the apache2 configuration
-if ! grep -qE "^. /etc/sysconfig/rudder-relay-apache$" /etc/sysconfig/apache2; then
+nextline=$(grep -A1 -E "^. /etc/sysconfig/rudder-relay-apache$" /etc/sysconfig/apache2 | tail -n1)
+if [ "${nextline}" = "" ]; then
+  # No include currently
   echo -e '# This sources the modules/defines needed by Rudder\n. /etc/sysconfig/rudder-relay-apache' >> /etc/sysconfig/apache2
+  echo -e '# This line is necessary for fillup not to remove any lines above. See #11153\nAPACHE_RUDDER_RELAY_CUSTOMIZED="true"' >> /etc/sysconfig/apache2
+elif [ "${nextline}" != "# This line is necessary for fillup not to remove any lines above. See #11153" ]; then
+  # Old include without comment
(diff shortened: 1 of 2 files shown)


rudder-server-relay/SPECS/rudder-server-relay.spec
```diff
@@ -318,10 +318,17 @@ fi
 
 %if 0%{?suse_version}
 # Add required includes in the apache2 configuration
-if ! grep -qE "^. /etc/sysconfig/rudder-relay-apache$" /etc/sysconfig/apache2; then
+nextline=$(grep -A1 -E "^. /etc/sysconfig/rudder-relay-apache$" /etc/sysconfig/apache2 | tail -n1)
+if [ "${nextline}" = "" ]; then
+  # No include currently
   echo -e '# This sources the modules/defines needed by Rudder\n. /etc/sysconfig/rudder-relay-apache' >> /etc/sysconfig/apache2
```
VinceMacBuche: Why don' we do 2 lines instead of one for each echo ? this makes it completely unreadable

VinceMacBuche: echo -e '# This sources the modules/defines needed by Rudder.' >> /etc/sysconfig/apache2 /etc/sysconfig/rudder-relay-apache'  >> /etc/sysconfig/apache2

rudder-server-relay/SPECS/rudder-server-relay.spec
```diff
@@ -318,10 +318,17 @@ fi
...
+  echo -e '# This line is necessary for fillup not to remove any lines above. See #11153\nAPACHE_RUDDER_RELAY_CUSTOMIZED="true"' >> /etc/sysconfig/apache2
+elif [ "${nextline}" != "# This line is necessary for fillup not to remove any lines above. See #11153" ]; then
+  # Old include without comment
+  sed -i 's|. /etc/sysconfig/rudder-relay-apache|. /etc/sysconfig/rudder-relay-apache\n# This line is necessary for fillup not to remove any lines above. See #11153\nAPACHE_RUDDER_RELAY_CUSTOMIZED="true"|' /etc/sysconfig/apache2
```
VinceMacBuche: Whye not the same echo as above ? and a complex sed ... because you need to have the line just below ?

amousset: yes, we want to insert the line just after the Rudder include.and keep a consistent file across versions.

amousset:
PR rebased

rudder-server-relay/SPECS/rudder-server-relay.spec
VinceMacBuche: ok!

PR #1513, opened 2018-02-28
Fixes #12162: On debian, installation is broken

amousset:
https://www.rudder-project.org/redmine/issues/12162


amousset:
according to http://www.fifi.org/doc/debconf-doc/tutorial.html#AEN34

amousset:
OK, merging this PR

PR #1514, opened 2018-02-28
Fixes #12083: On debian, with systemd rudder-agent services are not enabled during postinst

amousset:
https://www.rudder-project.org/redmine/issues/12083


amousset:
OK, merging this PR

PR #1517, opened 2018-03-02
Fixes #12186: Embed curl in the agent

peckpeck:
https://www.rudder-project.org/redmine/issues/12186


peckpeck:
Not terminated, we probably need one line somewhere in the dpkg/rpm part
We need to handle openssl properly (system vs embedded)
The interesting part would be to use the lib from within cfengine build
I don't know if we'll have time to merge and test this before 4.3

peckpeck:
Commit modified

peckpeck:
dpkg/rpm done
cfengine lib dine
openssl done

-> do we want this for 4.3 ?

PR #1519, opened 2018-03-12
Fixes #12204: After install, agent services are not enabled on Debian

VinceMacBuche:
https://www.rudder-project.org/redmine/issues/12204


VinceMacBuche:
OK, merging this PR

PR #1521, opened 2018-03-14
Fixes #12211: On debian with systemd, rudder-agent install fails with failed to read file: /var/rudder/cfengine-community/policy_server.dat

VinceMacBuche:
https://www.rudder-project.org/redmine/issues/12211


fanf:
OK, merging this PR

PR #1523, opened 2018-03-19
Fixes #12239: Epoch should be fixed in packaging

peckpeck:
https://www.rudder-project.org/redmine/issues/12239


peckpeck:
OK, merging this PR

PR #1540, opened 2018-03-28
Fixes #12298: Change defaults for embedding xml and yaml

peckpeck:
https://www.rudder-project.org/redmine/issues/12298


peckpeck:
Commit modified

PR #1575, opened 2018-04-10
Fixes #12400: Update rudder-agent-thin patch for new dependency

amousset:
https://www.rudder-project.org/redmine/issues/12400


amousset:
OK, merging this PR

PR #1577, opened 2018-04-11
Fixes #12406: Missing libattr package on rhel3 builder

amousset:
https://www.rudder-project.org/redmine/issues/12406


peckpeck:
Grr still the same mistake

PR #1581, opened 2018-04-16
Fixes #12449: Cleanup agent makefile

peckpeck:
https://www.rudder-project.org/redmine/issues/12449


rudder-agent/SOURCES/Makefile
```diff
@@ -139,7 +139,7 @@ endif
 ifneq (true,$(USE_SYSTEM_PERL))
 BUILD_DEPS := $(BUILD_DEPS) build-perl
 FUSION_DEPS := install-perl
-PERL:=$(INSTALL_DIR)/bin/perl
+PERL:=$(DESTDIR)/opt/rudder/bin/perl
```
amousset: why this change?

amousset: why not removing INSTALL_DIR completely?

rudder-agent/SOURCES/Makefile
```diff
@@ -199,10 +199,10 @@ build: get-flags build-cfengine
...
 	# Directories
-	mkdir -p $(INSTALL_DIR)/etc
-	mkdir -p $(INSTALL_DIR)/bin
-	mkdir -p $(INSTALL_DIR)/share/commands
-	mkdir -p $(INSTALL_DIR)/share/lib
+	mkdir -p $(DESTDIR)/opt/rudder/etc
+	mkdir -p $(DESTDIR)/opt/rudder/bin
+	mkdir -p $(DESTDIR)/opt/rudder/share/commands
+	mkdir -p $(DESTDIR)/opt/rudder/share/lib
 	mkdir -p $(DESTDIR)/var/rudder/cfengine-community/bin
```
peckpeck: The reason if this change can be seen here, mixing a little bit of INSTALL_DIR and DESTDIR everywhere make it difficult to read and understand where the final destination is

rudder-agent/SOURCES/Makefile
amousset: yes but it is still define, shouldn't we remove it completely

peckpeck:
Commit modified

PR #1590, opened 2018-05-04
Fixes #12583: Error during postinst on upgrade to 4.3

peckpeck:
https://www.rudder-project.org/redmine/issues/12583


peckpeck:
Commit modified

PR #1593, opened 2018-05-17
Fixes #12645: Build fail because on non existent man3 directory

VinceMacBuche:
https://www.rudder-project.org/redmine/issues/12645


VinceMacBuche:
Commit modified

PR #1596, opened 2018-05-18
Fixes #12641: Use pylint to verify python code in relay

peckpeck:
https://www.rudder-project.org/redmine/issues/12641


peckpeck:
Commit modified

PR #1598, opened 2018-05-24
Fixes #12634: .pyc file are not cleand up during postinst

amousset:
https://www.rudder-project.org/redmine/issues/12634

### DIFF
--- a/ncf-api-virtualenv/SPECS/ncf-api-virtualenv.spec
+++ b/ncf-api-virtualenv/SPECS/ncf-api-virtualenv.spec
@@ -247,6 +247,10 @@ fi
 # Enable mod_wsgi using a2enmod
 a2enmod wsgi >/dev/null 2>&1
 
+# Remove .pyc files to ensure we don't end up with outdated files
+rm -f %{installdir}/tools/ncf.pyc
+rm -f %{installdir}/tools/ncf_constraints.pyc
+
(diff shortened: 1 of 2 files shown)


amousset:
Commit modified

ncf-api-virtualenv/SPECS/ncf-api-virtualenv.spec
```diff
@@ -247,6 +247,10 @@ fi
 # Enable mod_wsgi using a2enmod
 a2enmod wsgi >/dev/null 2>&1
 
+# Remove .pyc files to ensure we don't end up with outdated files
+rm -f %{installdir}/tools/ncf.pyc
+rm -f %{installdir}/tools/ncf_constraints.pyc
```
VinceMacBuche: you should also rm ncf_rudder.pyc

VinceMacBuche: (in fact i would do 

> find %{installdir}/tools -name "*.pyc" -exec rm -f {} \\;

VinceMacBuche: Hmm, in fact ncf_rudder does not make a pyc ..., i would still do a find :)

peckpeck:
We should create another issue to make sure we never forget a .pyc file in any packages.
For today, this is sufficient I think

PR #1607, opened 2018-06-20
Fixes #7777: Rudder-agent provides rudder-agent without a version

amousset:
https://www.rudder-project.org/redmine/issues/7777


amousset:
Commit modified

PR #1608, opened 2018-06-25
Fixes #12831: Remove documentation from the install of rudder-agent dependencies in rudder-packages

VictorQrt:
https://www.rudder-project.org/redmine/issues/12831


peckpeck:
good but you need a rebase

PR #1610, opened 2018-07-11
Fixes #12331: rudder-pkg must disable plugin during major version Rudder update

fanf:
https://www.rudder-project.org/redmine/issues/12331

Technically, we don't need "plugin disable-all" and "plugin enable-all", but they are handy when managing a major upgrade (ie: in the documentation, we could add a disable-all before major upgrade to be on the safe side)


fanf:
PR rebased

PR #1611, opened 2018-07-11
Fixes #12930: During an upgrade of Rudder, if a new generic method appears in ncf, that is also on local method, all is broken

amousset:
https://www.rudder-project.org/redmine/issues/12930


amousset:
Commit modified

amousset:
Commit modified

amousset:
Commit modified

amousset:
Commit modified

amousset:
OK, merging this PR

PR #1620, opened 2018-07-17
Fixes #12976: Parent fix is not applied to the right place

amousset:
https://www.rudder-project.org/redmine/issues/12976


amousset:
Commit modified

PR #1626, opened 2018-07-20
Fixes #10646: Error at the end of generation after accepting 500 nodes

fanf:
https://www.rudder-project.org/redmine/issues/10646


ncharles:
init script should be updated, so that the value appears after upgrade. I'm taking over this pr

ncharles:
replaced by https://github.com/Normation/rudder-packages/pull/1627

PR #1629, opened 2018-07-25
Fixes #13041: rudder-metrics-reporting fail to report reports on directive/expected reports in 4.3, as the database schema changed

ncharles:
https://www.rudder-project.org/redmine/issues/13041

### DIFF
--- a/rudder-webapp/SOURCES/rudder-metrics-reporting
+++ b/rudder-webapp/SOURCES/rudder-metrics-reporting
@@ -54,6 +54,9 @@ import uuid
 def metrics():
   data = { }
 
+  # Define the versionning of the file
+  data['file_version'] = "2"
+
   # find uuid, or generate it
   data["uuid"] = get_uuid()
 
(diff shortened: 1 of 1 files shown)


rudder-webapp/SOURCES/rudder-metrics-reporting
```diff
@@ -143,6 +146,14 @@ def metrics():
   cmd="[ -x /usr/bin/dpkg ] && dpkg-query -W rudder-server-root || rpm -q rudder-server-root"
   data["package_version"] = command(["/bin/sh", "-c", cmd]).strip()
 
+  # Report mode
+  cmd="grep 'RUDDER_REPORT_MODE' /var/rudder/cfengine-community/inputs/rudder.json | sed 's/.*"RUDDER_REPORT_MODE":"\(.*\)",.*/\1/'"
```
peckpeck: this regex work with current json, but not necessarily if formatting change
I would remove the coma and add optional spaces around the semicolon

amousset: should we load rudder.json as a python object instead?

ncharles: this is exactly the regexp used in rudder agent info. 

@amousset do you thinkg we would really benefit from reading the json ?

ncharles: ok

ncharles:
Commit modified

PR #1633, opened 2018-07-26
Fixes #13063: Relay API does not correctly work on relay, as RewriteEngine is not enabled

ncharles:
https://www.rudder-project.org/redmine/issues/13063


ncharles:
OK, merging this PR

PR #1637, opened 2018-07-31
Fixes #13114: Use antora doc in rudder-webapp build

amousset:
https://www.rudder-project.org/redmine/issues/13114


amousset:
Commit modified

PR #1647, opened 2018-08-15
Purge of rudder-webapp should not fail

ludovicc:
Fixed the purge of package rudder-webapp which had the following issues:

- user rudder was left on the system
- attempt to delete group 'rudder' failed as a user (rudder) was still using it

### DIFF
--- a/rudder-webapp/debian/postrm
+++ b/rudder-webapp/debian/postrm
@@ -21,12 +21,16 @@ set -e
 
 case "$1" in
     purge)
+      # ignore errors during purge
+      set +e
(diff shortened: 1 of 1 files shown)


peckpeck:
Hello, thank you for the fix.

I forgot to tell you that this was fixed in https://github.com/Normation/rudder-packages/pull/1648 after opening the issue in redmine https://issues.rudder.io/issues/13210 which was released in Rudder 4.1.15, 4.3.5 and 5.0.1

It is my mistake that the commit is not attributed to you, sorry.

PR #1675, opened 2018-09-07
Fixes #12714: rudder syslog port 514 is reset after upgrade

ncharles:
https://www.rudder-project.org/redmine/issues/12714


amousset:
OK, merging this PR

PR #1676, opened 2018-09-11
Fixes #13437: relay package is missing a gpg dependency

peckpeck:
https://www.rudder-project.org/redmine/issues/13437

### DIFF
--- a/rudder-server-relay/SPECS/rudder-server-relay.spec
+++ b/rudder-server-relay/SPECS/rudder-server-relay.spec
@@ -98,7 +98,7 @@ Requires: rudder-agent >= %{real_epoch}:%{real_version}, rsyslog, openssl, %{apa
 
 ## RHEL
 %if 0%{?rhel}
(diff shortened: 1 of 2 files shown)


rudder-server-relay/SPECS/rudder-server-relay.spec
```diff
@@ -98,7 +98,7 @@ Requires: rudder-agent >= %{real_epoch}:%{real_version}, rsyslog, openssl, %{apa
 
 ## RHEL
 %if 0%{?rhel}
-Requires: mod_ssl
+Requires: mod_ssl, gnupg2
```
amousset: where is it used?

peckpeck: in the relay api

peckpeck: Well not sure

PR #1680, opened 2018-09-18
Fixes #13499: Doc download in webapp package does not work anymore

amousset:
https://www.rudder-project.org/redmine/issues/13499


amousset:
OK, merging this PR

PR #1687, opened 2018-09-21
Fixes #13549: rudder-agent service is not removed when removing rudder-agent

amousset:
https://www.rudder-project.org/redmine/issues/13549


amousset:
Commit modified

PR #1693, opened 2018-10-11
Fixes #13174: can't upgrade debian package if /bin/systemd exist but sysv is used

amousset:
https://www.rudder-project.org/redmine/issues/13174

### DIFF
--- a/rudder-agent/debian/postinst
+++ b/rudder-agent/debian/postinst
@@ -36,7 +36,7 @@ case "$1" in
 
     CFRUDDER_USE_SYSTEMD="false"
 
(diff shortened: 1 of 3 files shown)


amousset:
I did not used the proposed solution but the one used by CFEngine for systemd detection, what do you think?

amousset:
(Full CFEngine solution is to follow the symlinks and check the command end in "/systemd")

peckpeck:
This pr is not mergeable anymore

amousset:
Commit modified

amousset:
@peck updated

PR #1694, opened 2018-10-15
Fixes #13659: rudder-doc embedeed doesn't work in Rudder 4.3.5 nightly

amousset:
https://www.rudder-project.org/redmine/issues/13659


amousset:
OK, merging this PR

PR #1707, opened 2018-11-05
Fixes #13774: rudder-cf-execd and rudder-cf-serverd are not enabled nor running after a server install on debian 8

amousset:
https://www.rudder-project.org/redmine/issues/13774


amousset:
Commit modified

PR #1709, opened 2018-11-08
Fixes #13690: Connection error between agents and servers using openssl 1.0.x <-> 1.1.0

peckpeck:
https://www.rudder-project.org/redmine/issues/13690


rudder-agent/debian/control
```diff
@@ -2,7 +2,7 @@ Source: rudder-agent
 Section: admin
 Priority: extra
 Maintainer: Rudder packaging team <rudder-packaging@rudder-project.org>
-Build-Depends: debhelper (>= 7), libssl-dev, bison, gcc, flex, autoconf, automake, libtool, libpcre3-dev, libpam0g-dev, ca-certificates, perl, lsb-release, libyaml-dev, libxml2-dev, libcurl-ssl-dev | libcurl4-openssl-dev, libacl1-dev
```
amousset: we could keep for for future versions where we will be able to use system openssl/curl

peckpeck: it will be the occasion to rethink it
it think this is one cause of the double ld link in debian

amousset: Ok!

PR #1716, opened 2018-11-13
Fixes #11715: When "policy-server.dat" is missing, generated inventory is not legit

peckpeck:
https://issues.rudder.io/issues/11715


peckpeck:
OK, merging this PR

PR #1753, opened 2018-12-31
Fixes #14059: Force http 1.1 when downloading curl

amousset:
https://issues.rudder.io/issues/14059


amousset:
OK, merging this PR

PR #1800, opened 2019-02-19
Fixes #14339: Do not update policy_server.dat in case of upgrade even if rudder host is defined

amousset:
https://issues.rudder.io/issues/14339


amousset:
Commit modified

PR #1807, opened 2019-03-01
Fixes #14391: Policy generation hook "60-trigger-node-update" only works on the first node of the list

fanf:
https://issues.rudder.io/issues/14391


amousset:
:heart: python

PR #1829, opened 2019-03-21
Fixes #14515: Relink aix binaries with absolute path

peckpeck:
https://issues.rudder.io/issues/14515


peckpeck:
Commit modified

PR #1839, opened 2019-04-02
Fixes #14575: We should not remove cf-monitord as it breaks the policies

amousset:
https://issues.rudder.io/issues/14575


amousset:
Commit modified

PR #1842, opened 2019-04-03
Fixes #14580: when installing rudder-server-root, it asks about running rudder-node-to-relay, which is misleading (and wrong)

ncharles:
https://issues.rudder.io/issues/14580


ncharles:
for the record, i ran manually the command at the begining of the install, and so i didn't have the message about relay, an not new error (even, less)

PR #1846, opened 2019-04-08
Fixes #14602: Agent stderr shows up in /var/log/apache2/error.log

ncharles:
https://issues.rudder.io/issues/14602

### DIFF
--- a/rudder-server-relay/SOURCES/relay-api/relay_api/remote_run.py
+++ b/rudder-server-relay/SOURCES/relay-api/relay_api/remote_run.py
@@ -23,7 +23,7 @@ except:
 
 NEXTHOP = None
 REMOTE_RUN_COMMAND = "sudo /opt/rudder/bin/rudder remote run"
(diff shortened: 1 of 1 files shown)


rudder-server-relay/SOURCES/relay-api/relay_api/remote_run.py
```diff
@@ -23,7 +23,7 @@
 
 NEXTHOP = None
 REMOTE_RUN_COMMAND = "sudo /opt/rudder/bin/rudder remote run"
-LOCAL_RUN_COMMAND = "sudo /opt/rudder/bin/rudder agent run"
+LOCAL_RUN_COMMAND = "sudo /opt/rudder/bin/rudder agent run &> /dev/null"
```
VinceMacBuche: ```&>``` is not available in all shell, and we should use ``` > /dev/nul 2>&1 ``` instead ... But maybe &> is available everywhere ? because it seems better to me ...

PR #1850, opened 2019-04-09
Fixes #14616: rudder agent start complains loudly at install due to missing /opt/rudder/etc/uuid.hive

ncharles:
https://issues.rudder.io/issues/14616


ncharles:
it first needs https://github.com/Normation/rudder-agent/pull/208

PR #1852, opened 2019-04-12
Fixes #14644: When installing rudder-agent, there's a long wait at the end of the install because of #14258

ncharles:
https://issues.rudder.io/issues/14644


ncharles:
OK, merging this PR

ncharles:
OK, merging this PR

PR #1856, opened 2019-04-16
Fixes #14671: Remove http1.1 parameter from curl

amousset:
https://issues.rudder.io/issues/14671


amousset:
OK, merging this PR

PR #1867, opened 2019-04-17
Fixes #14696: relayd fails during package creation on RPM 

amousset:
https://issues.rudder.io/issues/14696


rudder-server-relay/SPECS/rudder-server-relay.spec
```diff
@@ -16,6 +16,9 @@
 #
 #####################################################################################
 
+# Disable totally debug info package
+%define debug_package %{nil}
```
peckpeck: shouldn't we put this everywhere ?

PR #1894, opened 2019-05-06
Fixes #14829: Error in parent ticket - missing Digest/SHA.pm as a dependency on centos

ncharles:
https://issues.rudder.io/issues/14829


ncharles:
OK, merging this PR

ncharles:
OK, merging this PR

PR #1898, opened 2019-05-07
Fixes #14833: Use rudder agent check at postinst and factory reset to avoid duplacing code

peckpeck:
https://issues.rudder.io/issues/14833


peckpeck:
Wrong issue

PR #1910, opened 2019-05-14
Fixes #14881: Move away from python 2 to python 3

peckpeck:
https://issues.rudder.io/issues/14881


rudder-webapp/SOURCES/rudder-metrics-reporting
```diff
@@ -1,6 +1,5 @@
-#!/usr/bin/python
+#!/usr/bin/python3
 
-from __future__ import print_function
```
Fdall: this is mandatory to make the script work in python2 for rhel7

PR #1911, opened 2019-05-14
Fixes #14887: Move away from python 2 to python 3 in virtualenv

peckpeck:
https://issues.rudder.io/issues/14887


peckpeck:
do not merge we must check system dependencies

PR #1929, opened 2019-05-24
Fixes #14945: Missing index in ldap directory 

ncharles:
https://issues.rudder.io/issues/14945

### DIFF
--- a/rudder-inventory-ldap/SOURCES/slapd.conf
+++ b/rudder-inventory-ldap/SOURCES/slapd.conf
@@ -66,6 +66,11 @@ maxsize 104857600
 
 # Indices to maintain
 index	objectClass	eq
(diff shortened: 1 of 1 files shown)


fanf:
I would like to check on load plateform these changes before merging them.

fanf:
I would like to check on load plateform these changes before merging them.

ncharles:
impact is totally null on load platform: neither better nor worse :/

fanf:
Should we merge that? If the impact if null, it seems like a risk to add them (we don't know the impact on write, which can be important). 
Perhaps we should just not do anything?

ncharles:
on some platform it's faster, but indeed the impact is too risky (or at least feels risky). Closing

PR #1954, opened 2019-06-03
Fixes #15013: dh_install is called with a wrong option in all builder 

Fdall:
https://issues.rudder.io/issues/15013


Fdall:
OK, merging this PR

PR #2024, opened 2019-08-06
Fixes #15417: rudder-webapp package cannot install, as the webapp exits when the ldap is not initialized 

ncharles:
https://issues.rudder.io/issues/15417


Fdall:
OK, merging this PR

PR #2038, opened 2019-08-21
Fixes #15542: Rudder-pkg doesn't bring its dependencies in the build depencies and fails to build

Fdall:
https://issues.rudder.io/issues/15542


Fdall:
PR updated with a new commit

Fdall:
OK, squash merging this PR

PR #2042, opened 2019-08-23
Fixes #15566: rudder-pkg is listed in the package files and is created in postinst so it can not work

Fdall:
https://issues.rudder.io/issues/15566


Fdall:
Commit modified

rudder-server-relay/SPECS/rudder-server-relay.spec
```diff
@@ -266,6 +266,7 @@ rm -rf %{buildroot}
 /var/rudder/reports/incoming
 /var/rudder/shared-files/
 /var/rudder/share/
+/var/rudder/share/commands/package
```
amousset: no need as ∕var/rudder/share is already part of the package

Fdall:
Closing since it is already ok

PR #2063, opened 2019-09-10
Fixes #10623: Hooks content/permissions are changed during Rudder upgrade (for ex  /opt/rudder/etc/hooks.d/policy-generation-node-ready/10-cf-promise-check is replaced)

peckpeck:
https://issues.rudder.io/issues/10623


peckpeck:
OK, merging this PR

PR #2088, opened 2019-09-19
Fixes #15763: Create /opt/rudder/etc/plugins/licenses at server install

VinceMacBuche:
https://issues.rudder.io/issues/15763


VinceMacBuche:
OK, merging this PR

PR #2116, opened 2019-10-10
Fixes #15912: Apply SELinux configuration in relay

amousset:
https://issues.rudder.io/issues/15912


amousset:
OK, merging this PR

PR #2122, opened 2019-10-16
Fixes #15979: Unwanted systemctl output in rudder-reports postinst

amousset:
https://issues.rudder.io/issues/15979


amousset:
OK, merging this PR

PR #2123, opened 2019-10-16
Fixes #15985: Remove debug pprint from rudder-pkg

amousset:
https://issues.rudder.io/issues/15985


amousset:
OK, merging this PR

PR #2146, opened 2019-11-13
Fixes #15188: Allow disabling the agent without stopping cf-serverd

peckpeck:
https://issues.rudder.io/issues/15188


peckpeck:
Commit modified

PR #2181, opened 2020-01-08
Fixes #16500: Upgrade from 5.0.x to 6.0.x on debian fail because of conflicting file

Fdall:
https://issues.rudder.io/issues/16500


Fdall:
Commit modified

PR #2182, opened 2020-01-09
Fixes #16509: on centos8 relay, service relayd is not started after installation

amousset:
https://issues.rudder.io/issues/16509


amousset:
Replaced by https://github.com/Normation/rudder-packages/pull/2183

PR #2189, opened 2020-01-15
Fixes #16559: Rpm packages are missing some conffiles since /opt/rudder/etc is not included as files

VinceMacBuche:
https://issues.rudder.io/issues/16559

### DIFF
--- a/rudder-server-relay/SPECS/rudder-server-relay.spec
+++ b/rudder-server-relay/SPECS/rudder-server-relay.spec
@@ -293,15 +293,18 @@ rm -rf %{buildroot}
 /var/rudder/lib/relay/
 /var/log/rudder/apache2/
 /usr/lib/systemd/system/rudder-relayd.service
+/opt/rudder/bin/relay-cleanup
 /opt/rudder/bin/rudder-relayd
+/opt/rudder/bin/rudder-pkg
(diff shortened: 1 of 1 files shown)


rudder-server-relay/SPECS/rudder-server-relay.spec
```diff
@@ -274,10 +274,13 @@ rm -rf %{buildroot}
 %config(noreplace) /etc/%{apache_vhost_dir}/rudder.conf
 %config(noreplace) /opt/rudder/etc/rudder-apache-relay-common.conf
 %config(noreplace) /opt/rudder/etc/rudder-apache-relay-ssl.conf
+%config(noreplace) /opt/rudder/etc/rudder-apache-relay-nossl.conf
 %config(noreplace) /opt/rudder/etc/rudder-networks-24.conf
 %config(noreplace) /opt/rudder/etc/rudder-networks-policy-server-24.conf
 %config(noreplace) /opt/rudder/etc/rudder-pkg/rudder-pkg.conf
+%config(noreplace) /opt/rudder/etc/rudder-pkg/rudder_plugins_key.pub
```
amousset: this one should not be in noreplace, but it's the same for apache config and openssl config at least...

VinceMacBuche:
Commit modified

PR #2198, opened 2020-02-11
Fixes #16707: After upgrade from 5.0 to 6.0 on centos7, rudder-slapd is not started

ncharles:
https://issues.rudder.io/issues/16707

### DIFF
--- a/rudder-webapp/SPECS/rudder-webapp.spec
+++ b/rudder-webapp/SPECS/rudder-webapp.spec
@@ -359,6 +359,14 @@ then
   systemctl stop rudder-slapd
 fi
 
+#=================================================
+# Post transaction
+#=================================================
+%posttrans -n rudder-webapp
+
+# during upgrade, service may have been stopped by uninstall of rudder-inventory-ldap
+systemctl start rudder-slapd
+
 
 #=================================================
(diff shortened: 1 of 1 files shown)


rudder-webapp/SPECS/rudder-webapp.spec
```diff
@@ -359,6 +359,24 @@ then
   systemctl stop rudder-slapd
 fi
 
+#=================================================
+# Post transaction
+#=================================================
+%posttrans -n rudder-webapp
+
```
amousset: We can just run `systemctl start rudder-slapd` anyway

ncharles:
PR updated with a new commit

rudder-webapp/SPECS/rudder-webapp.spec
```diff
@@ -359,6 +359,14 @@ then
...
+# during upgrade, service may have been stopped by uninstall of rudder-inventory-ldap
+systemctl start rudder-slapd
```
amousset:
```suggestion
systemctl start rudder-slapd >/dev/null
```

ncharles:
OK, squash merging this PR

PR #2208, opened 2020-02-23
Fixes #16790: Our agent package contains CFEngine systemd units

amousset:
https://issues.rudder.io/issues/16790

### DIFF
--- a/rudder-agent/SOURCES/Makefile
+++ b/rudder-agent/SOURCES/Makefile
@@ -666,6 +666,8 @@ install-cfengine: build-cfengine $(INSTALL_DEPS) rudder.8.gz
 		LD_LIBRARY_PATH="$(DESTDIR)/opt/rudder/lib" ./$${binary} -M | gzip > $(DESTDIR)/opt/rudder/share/man/man8/$${binary}.8.gz; \
 	done
 	cd $(DESTDIR)/opt/rudder && rm -rf share/doc/cfengine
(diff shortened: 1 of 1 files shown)


rudder-agent/SOURCES/Makefile
```diff
@@ -649,6 +649,8 @@ install-cfengine: build-cfengine $(INSTALL_DEPS) rudder.8.gz
 		LD_LIBRARY_PATH="$(DESTDIR)/opt/rudder/lib" ./$${binary} -M | gzip > $(DESTDIR)/opt/rudder/share/man/man8/$${binary}.8.gz; \
 	done
 	cd $(DESTDIR)/opt/rudder && rm -rf share/doc/cfengine
+	# Remove CFEngine systemd units (ours are in /lib/systemd)
+	cd $(DESTDIR) && rm -rf usr/lib/systemd/system/
```
peckpeck: what if other systemd units have be installed in DESTDIR before ?
can't you remove files explicitly ?

amousset: We don't use the same dir for systemd units, but we can name them indeed

peckpeck: but other products in DESTDIR may do

amousset:
PR updated with a new commit

amousset:
@peck updated

amousset:
OK, squash merging this PR

PR #2212, opened 2020-03-02
Fixes #16845: Policy generation fails with many nodes in 6.0 after an upgrade

ncharles:
https://issues.rudder.io/issues/16845


rudder-webapp/SOURCES/rudder-upgrade
```diff
@@ -526,6 +527,15 @@ propertyName: rudder_report_protocol_default
 propertyValue: SYSLOG
 EOF
   fi
+
+  # 6.0.4:
+  # ensure sockbuf_max_incoming_auth is set in /opt/rudder/etc/openldap/slapd.conf
+  MAX_INCOMING_AUTH_SET=$(grep -c "sockbuf_max_incoming_auth" /opt/rudder/etc/openldap/slapd.conf || true)
```
amousset:
```suggestion
  MAX_INCOMING_AUTH_SET=$(grep -c "sockbuf_max_incoming_auth" /opt/rudder/etc/openldap/slapd.conf 2>/dev/null || true)  
```

PR #2216, opened 2020-03-06
Fixes #16889: Standardize os name in agent postinst call

peckpeck:
https://issues.rudder.io/issues/16889

### DIFF
--- a/rudder-agent/SPECS/rudder-agent.spec
+++ b/rudder-agent/SPECS/rudder-agent.spec
@@ -76,7 +76,7 @@
 %endif
 
 # OS family to build for
-%if "%{?_os}" == "aix"
+%if 0%{?aix}
 %define os_family aix
 %else
 %define os_family linux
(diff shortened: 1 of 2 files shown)


rudder-agent/SPECS/rudder-agent.spec
```diff
@@ -387,7 +387,7 @@ then
   CFRUDDER_FIRST_INSTALL="true"
 fi
 
-/opt/rudder/share/package-scripts/rudder-agent-postinst "${CFRUDDER_FIRST_INSTALL}" "%{?_os}" "%{use_systemd}" ""
+/opt/rudder/share/package-scripts/rudder-agent-postinst "${CFRUDDER_FIRST_INSTALL}" "rpm" "%{use_systemd}" ""
```
Fdall: Why rpm here ?

Fdall: even if the option is currently ignored, rpm is confusing

Fdall: Ok, actually we are already using this syntax on the debian build so it is coherent.

PR #2222, opened 2020-03-19
Fixes #14600: when installing rudder-server-root, it asks about running rudder-node-to-relay, on Centos and Sles

ncharles:
https://issues.rudder.io/issues/14600


ncharles:
OK, merging this PR

ncharles:
OK, merging this PR

PR #2230, opened 2020-04-15
Fixes #17121: Make agents preserve perm on copy to keep previous behavior

amousset:
https://issues.rudder.io/issues/17121

### DIFF
--- a/rudder-agent/SOURCES/patches/cfengine/17121-preserve-perms-copy.patch
+++ b/rudder-agent/SOURCES/patches/cfengine/17121-preserve-perms-copy.patch
@@ -1,0 +1,15 @@
+diff --git a/cf-agent/verify_files_utils.c b/cf-agent/verify_files_utils.c
+index 7046026..383c69b 100644
+--- a/cf-agent/verify_files_utils.c
++++ b/cf-agent/verify_files_utils.c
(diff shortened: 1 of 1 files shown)


rudder-agent/SOURCES/patches/cfengine/17121-preserve-perms-copy.patch
```diff
@@ -0,0 +1,15 @@
...
+--- a/cf-agent/verify_files_utils.c
++++ b/cf-agent/verify_files_utils.c
+@@ -1366,7 +1366,9 @@ bool CopyRegularFile(EvalContext *ctx, const char *source, const char *dest, str
+     else
+     {
+         // If preserve is true, retain permissions of source file
+-        if (attr.copy.preserve)
++        //if (attr.copy.preserve)
++	// Keep previous behavior to stay compatible
++	if (true)
```
peckpeck, line 12: shouldn't we allow someone to put false here ? 
this doesn't look like the default value, but every value

amousset: indeed,but that was the previous behavior. There is no way to know if it is an explicit false or a default false.

PR #2249, opened 2020-04-24
Fixes #17128: review index for LDAP

ncharles:
https://issues.rudder.io/issues/17128

### DIFF
--- a/rudder-webapp/SOURCES/slapd.conf
+++ b/rudder-webapp/SOURCES/slapd.conf
@@ -61,6 +61,9 @@ maxsize 107374182400
 
 # Indices to maintain
 index	objectClass	eq
+index	cn		eq
+index	softwareVersion	eq
+index	directiveId	eq
 
 # 2 - Monitor
 # Query statistics here by running:


rudder-webapp/SOURCES/slapd.conf
```diff
@@ -61,6 +61,9 @@ maxsize 107374182400
 
 # Indices to maintain
 index	objectClass	eq
+index	cn		eq
+index	softwareVersion	eq
+index	directiveId	eq
```
fanf, line 67: Because this subject is hard to check but in the long run, I propose to add in comments why/how each elements were tested (and especially if the were not). Ex:
```
# from https://issues.rudder.io/issues/17128
index	cn		eq
index	softwareVersion	eq
index	directiveId	eq
# this one leads to an error, not explored yet: https://issues.rudder.io/issues/17128#note-8
# index nodeGroupId eq #extremelly frequent!
# appear in candidate but not tested yet, sorted by number, more in top:
# index nodeId eq 
# index osName eq
# index agentName eq
# index isDynamic eq
# index isSystem eq
# index policyServerId eq
# index software eq
```

PR #2274, opened 2020-05-07
Fixes #17215: create migration script for LDAP indexes

ncharles:
https://issues.rudder.io/issues/17215

### DIFF
--- a/rudder-webapp/SOURCES/rudder-upgrade
+++ b/rudder-webapp/SOURCES/rudder-upgrade
@@ -53,6 +53,7 @@ trap 'anomaly_handler ${LINENO}' ERR INT TERM
 # - 6.0.0        : Keep syslog as default ptotocol on upgrade
 # - 6.0.4        : Ensure that rudder-slapd sockbuf_max_incoming_auth is not lost at upgrade 
 # - 6.1.0        : Migrate RudderUniqueID to the template format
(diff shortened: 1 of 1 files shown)


rudder-webapp/SOURCES/rudder-upgrade
```diff
@@ -549,7 +557,33 @@ EOF
...
+    line="index\s*${i}\s*eq"
+    INDEX_COUNT=$(grep -c "^${line}" ${SLAPD_CONF} 2>/dev/null || true)
+    if [ ${INDEX_COUNT} -eq 0 ]; then
+      echo "Adding LDAP index on attribute: ${i}"
+      sed -i "/index\s*objectClass\s*eq/a index\t${i}\teq" ${SLAPD_CONF}
+      NEED_REINDEX=true
+    fi
+  done
+
+  if ${NEED_REINDEX}; then
```
peckpeck, line 575: almost everything is a text in so we need comparison in tests : if [ "${NEED_REINDEX}" = "true" ]; then

ncharles: all other comparision are that way, like 
`if ${RUDDER_DB_INITIALIZED}; then`

so that would be inconsistent with the rest of the script

ncharles:
OK, squash merging this PR

PR #2281, opened 2020-05-09
Fixes #17365: strip calls fail on aix

peckpeck:
https://issues.rudder.io/issues/17365


peckpeck:
I don't know hos it would have worked before

PR #2286, opened 2020-05-12
Fixes #17386: We use the "service" command (on CentOS7) but do not not depend on the package providing it

amousset:
https://issues.rudder.io/issues/17386

### DIFF
--- a/rudder-webapp/SOURCES/rudder-webapp-postinst
+++ b/rudder-webapp/SOURCES/rudder-webapp-postinst
@@ -90,27 +90,27 @@ fi
 # Need to restart to take schema changes into account
 echo -n "INFO: Restarting rudder-slapd..."
 
-# during upgrade, service may not have been started with systemctl
-# detect first if rudder-slapd is seen running ?
-RUDDER_SLAPD_RUNNING_SERVICE=0
-RUDDER_SLAPD_RUNNING_SYSTEMCTL=0
-if service rudder-slapd status > /dev/null; then 
-  RUDDER_SLAPD_RUNNING_SERVICE=1
-fi
-if systemctl is-active rudder-slapd > /dev/null; then 
-  RUDDER_SLAPD_RUNNING_SYSTEMCTL=1
-fi
+if type service >/dev/null 2>&1; then
+  # during upgrade, service may not have been started with systemctl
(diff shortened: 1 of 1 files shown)


rudder-webapp/SOURCES/rudder-webapp-postinst
```diff
@@ -90,28 +90,28 @@ fi
...
-fi
-if systemctl is-active rudder-slapd > /dev/null; then 
-  RUDDER_SLAPD_RUNNING_SYSTEMCTL=1
-fi
-
-if [ $RUDDER_SLAPD_RUNNING_SERVICE -eq 1 ] && [ $RUDDER_SLAPD_RUNNING_SYSTEMCTL -eq 0 ]
-then
-  # systemctl doesn't know the service is started, we stop it with service first
-  service rudder-slapd stop >> ${LOG_FILE}
+if type service >/dev/null 2>&1; then
```
peckpeck, line 93: Shouldn't we do something if it doesn't exist ?

rudder-webapp/SOURCES/rudder-webapp-postinst
```diff
@@ -90,28 +90,28 @@ fi
...
+  if systemctl is-active rudder-slapd > /dev/null; then 
+    RUDDER_SLAPD_RUNNING_SYSTEMCTL=1
+  fi
+
+  if [ $RUDDER_SLAPD_RUNNING_SERVICE -eq 1 ] && [ $RUDDER_SLAPD_RUNNING_SYSTEMCTL -eq 0 ]
+  then
+    # systemctl doesn't know the service is started, we stop it with service first
+    service rudder-slapd stop >> ${LOG_FILE}
+  fi
 fi
```
peckpeck, line 110: Since we use systemctl a fe lines above, shouldn't we use it everywhere ?

amousset: we use it everywhere except for this specific migration code

rudder-webapp/SOURCES/rudder-webapp-postinst
amousset: I don't know, because it is very unlike the service was started without systemd if the system does not have sysvinit installed

PR #2287, opened 2020-05-13
Fixes #17395: SELinux policy for technique editor is not applied anymore after upgrade on RHEL/Centos server

VinceMacBuche:
https://issues.rudder.io/issues/17395


rudder-webapp/SPECS/rudder-webapp.spec
```diff
@@ -370,6 +370,15 @@ fi
 systemctl start rudder-slapd >/dev/null
 systemctl start rudder-jetty >/dev/null
 
+%if 0%{?rhel}
```
ncharles, line 373: wouldn't it be better if it were before starting the services?

PR #2290, opened 2020-05-18
Fixes #17429: Warn users if scale-out-relay plugin is not installed when promoting node through script

ElaadF:
https://issues.rudder.io/issues/17429

### DIFF
--- a/rudder-webapp/SOURCES/rudder-node-to-relay
+++ b/rudder-webapp/SOURCES/rudder-node-to-relay
@@ -36,4 +36,10 @@
 
 echo "INFO: This command is deprecated, please use 'rudder server node-to-relay <UUID>'"
 echo -e "Running: rudder server node-to-relay $@ \n"
-rudder server node-to-relay $@
+
+if ! [ "$(rudder package list | grep -q rudder-plugin-scale-out-relay)" ]; then
+  echo -e 'Error: This command requires `scale-out-relay` plugin' >&2
+  exit 1
+else
+  rudder server node-to-relay $@
+fi


rudder-webapp/SOURCES/rudder-node-to-relay
```diff
@@ -36,4 +36,10 @@
 
 echo "INFO: This command is deprecated, please use 'rudder server node-to-relay <UUID>'"
 echo -e "Running: rudder server node-to-relay $@ \n"
-rudder server node-to-relay $@
+
+if ! [ "$(command -v rudder server node-to-relay >/dev/null)" ]; then
```
amousset, line 40: What is this supposed to do? To detect the plugin presence it's probably better to query `rudder package` installed plugins list.

fanf: since we redirect to the command, it's likely what we want to test here. Not sure if it's the correct way to do it, but it seems like it's the correct thing to test.

ElaadF:
Commit modified

rudder-webapp/SOURCES/rudder-node-to-relay
```diff
@@ -36,4 +36,10 @@
 
 echo "INFO: This command is deprecated, please use 'rudder server node-to-relay <UUID>'"
 echo -e "Running: rudder server node-to-relay $@ \n"
-rudder server node-to-relay $@
+
+if ! [ "$(rudder package list | grep rudder-plugin-scale-out-relay)" ]; then
```
amousset, line 40: `grep -q` to avoid displaying the output

ElaadF:
Commit modified

PR #2306, opened 2020-06-02
Fixes #17592: rudder-webapp could suggest rudder-reports

peckpeck:
https://issues.rudder.io/issues/17592


rudder-webapp/debian/control
```diff
@@ -12,6 +12,7 @@ Pre-Depends: ${rudder:deps}
 Depends: ${shlibs:Depends}, ${misc:Depends}, systemd, iproute2, rsyslog, python3, apache2, apache2-utils, git-core, rsync, lsb-release, openssl, ldap-utils, postgresql-client (>=9.2), openjdk-11-jre-headless | openjdk-8-jre-headless | oracle-java8-installer | java11-runtime-headless, curl, acl
 Breaks: ncf, ncf-api-virtualenv, rudder-techniques, rudder-inventory-endpoint, rudder-jetty, rudder-inventory-ldap, rudder-ldap, rudder-agent (<< 6.0.0)
 Replaces: ncf, ncf-api-virtualenv, rudder-techniques, rudder-inventory-endpoint, rudder-jetty, rudder-inventory-ldap, rudder-ldap
+Suggests: rudder-reports
```
ncharles, line 15: i think it should Recommends
https://www.debian.org/doc/debian-policy/ch-relationships.html

> Recommends

    This declares a strong, but not absolute, dependency.

    The Recommends field should list packages that would be found together with this one in all but unusual installations.
Suggests

    This is used to declare that one package may be more useful with one or more others. Using this field tells the packaging system and the user that the listed packages are related to this one and can perhaps enhance its usefulness, but that installing this one without them is perfectly reasonable.

PR #2308, opened 2020-06-04
Fixes #17606: Agent uses "cmp" command but it is not a dependency

amousset:
https://issues.rudder.io/issues/17606


amousset:
Commit modified

amousset:
Commit modified

amousset:
We use it to compare two files:

```
if cmp -s "/var/rudder/cfengine-community/inputs/failsafe.cf" "/opt/rudder/share/bootstrap-promises/failsafe.cf"
```

we can probably use a tool we already have.

PR #2315, opened 2020-06-10
Fixes #17695: /opt/rudder/etc/rudder-pkg comes with the wrong permissions

Fdall:
https://issues.rudder.io/issues/17695


peckpeck:
what about debian ?

Fdall:
> what about debian ?

See here: https://github.com/Normation/rudder/pull/3072

PR #2318, opened 2020-06-11
Fixes #17717: systemd detection fails when init has parameters

peckpeck:
https://issues.rudder.io/issues/17717

### DIFF
--- a/rudder-agent/debian/postinst
+++ b/rudder-agent/debian/postinst
@@ -32,7 +32,7 @@ case "$1" in
 
     CFRUDDER_USE_SYSTEMD="false"
 
-    if readlink -f $(cut -f 1 -d "" < /proc/1/cmdline) | grep -qE "/systemd$" ; then CFRUDDER_USE_SYSTEMD="true"; fi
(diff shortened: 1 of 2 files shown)


amousset:
This produces a `bash: warning: command substitution: ignored null byte in input
` when systemd has no parameters

amousset:
Plus it does not work as `/usr/lib/systemd/systemd--switched-root--system--deserialize31` does not match `/systemd$`

peckpeck:
PR updated with a new commit

rudder-agent/debian/postinst
```diff
@@ -32,7 +32,7 @@ case "$1" in
 
     CFRUDDER_USE_SYSTEMD="false"
 
-    if readlink -f $(cut -f 1 -d " " < /proc/1/cmdline) | grep -qE "/systemd$" ; then CFRUDDER_USE_SYSTEMD="true"; fi
+    if readlink -f $(strings /proc/1/cmdline | head -n1 | cut -f 1 -d " ") | grep -qE "/systemd$" ; then CFRUDDER_USE_SYSTEMD="true"; fi
```
amousset, line 35: this requires to add binutils as dependency for the agent

amousset: it is not installed by default in base docker images, and probably some stripped-down VMs too.

peckpeck:
PR updated with a new commit

peckpeck:
OK, squash merging this PR

PR #2329, opened 2020-07-14
Fixes #17958: Create a configure to make the agent package -  configure target doesn't exist

peckpeck:
https://issues.rudder.io/issues/17958

### DIFF
--- a/rudder-agent/debian/rules
+++ b/rudder-agent/debian/rules
@@ -103,12 +103,9 @@ ifeq ($(findstring systemd,$(WITH)),)
 DEPS := $(DEPS), systemd
 endif
 
-configure:
-	dh_testdir
-	cd SOURCES && ./configure $(WITH)
-
 build:
 	dh_testdir
+	cd SOURCES && ./configure $(WITH)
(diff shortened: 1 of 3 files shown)


peckpeck:
PR updated with a new commit

rudder-agent/debian/rules
```diff
@@ -103,12 +103,9 @@ ifeq ($(findstring systemd,$(WITH)),)
 DEPS := $(DEPS), systemd
 endif
 
-configure:
-	dh_testdir
-	cd SOURCES && ./configure $(WITH)
-
 build:
 	dh_testdir
+	cd SOURCES && ./configure $(WITH)
```
amousset, line 108: why?

peckpeck: why not ? :)

peckpeck: That's the point of the parent PR : having a configure to call before calling make

amousset: indeed :D but why moving the configure from the configure target to build?

peckpeck: I don't know why the configre target was there first but it doesn't exist in debian specification and thus is never called

amousset: :man_shrugging:

peckpeck:
OK, squash merging this PR

PR #2376, opened 2020-10-27
Fixes #18435: Version rudder-api-client like Rudder

amousset:
https://issues.rudder.io/issues/18435


peckpeck:
This won't be sufficient :)

amousset:
I think it will.

peckpeck:
oh it was only on rpm that it didn't work !

peckpeck (approved):

PR #2407, opened 2020-12-04
Fixes #18742: Remove plugin notice during upgrade

VinceMacBuche:
https://issues.rudder.io/issues/18742


amousset:
Ok, for me, waiting for @Fdall approval too

fanf (approved):

fanf:
It's ok with me too, and I would to be able to test it and know that we are ready for the final today

PR #2415, opened 2021-02-14
Fixes #13812: Improve hardening flags for our C dependencies

amousset:
https://issues.rudder.io/issues/13812

### DIFF
--- a/rudder-agent/SOURCES/configure
+++ b/rudder-agent/SOURCES/configure
@@ -212,22 +212,37 @@ fi
 test_cc() {
   TEST_FILE=$(mktemp /tmp/cc.XXXXXX)
   echo "void main() {}" > "${TEST_FILE}.c"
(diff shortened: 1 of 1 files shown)


rudder-agent/SOURCES/configure
```diff
@@ -212,23 +212,38 @@ fi
 test_cc() {
   TEST_FILE=$(mktemp /tmp/cc.XXXXXX)
   echo "void main() {}" > "${TEST_FILE}.c"
-  if cc -o "$2" "$3" "${TEST_FILE}" "${TEST_FILE}.c" 2>/dev/null
```
amousset, line 215: did not work as expected with several options (but still returned zero so it was always enabled):

```
+ cc -o -fstack-protector '-Wl,-z,relro -Wl,-z,now' /tmp/cc.N2CQ1s /tmp/cc.N2CQ1s.c
/usr/bin/ld: warning: -z relro -Wl ignored
```

peckpeck (approved):

PR #2459, opened 2021-03-13
Fixes #19027: fail to build on solaris

peckpeck:
https://issues.rudder.io/issues/19027

### DIFF
--- a/rudder-agent/SOURCES/configure
+++ b/rudder-agent/SOURCES/configure
@@ -225,11 +225,11 @@ test_cc() {
 # test options one by one to enable as many as possible
 
 # try strong first
-test_cc "SECURE" "-fstack-protector-strong" ""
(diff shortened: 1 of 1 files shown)


rudder-agent/SOURCES/configure
```diff
@@ -225,11 +225,11 @@ test_cc() {
 # test options one by one to enable as many as possible
 
 # try strong first
-test_cc "SECURE" "-fstack-protector-strong" ""
+test_cc "SECURE" "-fstack-protector-strong" "-fstack-protector-strong"
```
amousset, line 228: why passing it as ldflag?

peckpeck: because security protections need various libraries and the linker knows which ones but sometimes doesn't guess by itself it should include them

PR #2481, opened 2021-05-12
Fixes #19261: Migration script for rudder.community.port

fanf:
https://issues.rudder.io/issues/19261

### DIFF
--- a/rudder-webapp/SOURCES/rudder-upgrade
+++ b/rudder-webapp/SOURCES/rudder-upgrade
@@ -52,6 +52,7 @@ trap 'anomaly_handler ${LINENO}' ERR INT TERM
 # - 6.1.1        : Update apache conf if it does not contain technique editor conf (file not updated by packaging
 # - 6.2.0        : Remove unused index on RudderSysEvents  
 # - 6.2.0~beta2  : Add sofware and modifyTimestamp indexes in slapd.conf
+# - 7.0.0~alpha1 : Rename property rudder.community.port into rudder.policy.distribution.port.cfengine
 ####################################################################################
 
 # Some paths
@@ -379,6 +380,21 @@ upgrade_masterfiles() {
 # Check and update config
 ################################################################################
 
+upgrade_conf_file() {
+
+  STEP="Upgrade /opt/rudder/etc/rudder-web.properties"
+  CONF_FILE="/opt/rudder/etc/rudder-web.properties"
+
+  # - 7.0.0~alpha1 : Rename property rudder.community.port into rudder.policy.distribution.port.cfengine
+  OLD_PROPERTY_NAME="rudder.community.port"
+  NEW_PROPERTY_NAME="rudder.policy.distribution.port.cfengine"
+  ATTRIBUTESET=$(grep -c "^${NEW_PROPERTY_NAME}[ \t]*=" ${CONF_FILE} || true)
+  if [ ${ATTRIBUTESET} -eq 0 ]; then
+    sed -i -e "s/${OLD_PROPERTY_NAME}/${NEW_PROPERTY_NAME}/" ${CONF_FILE}
+    echo "INFO: Configuration property '${OLD_PROPERTY_NAME}' was updated to '${NEW_PROPERTY_NAME}' in /opt/rudder/etc/rudder-web.properties"
+  fi
+
+}
 
 ################################################################################
 # Check and update database
@@ -562,6 +578,7 @@ upgrade_plugins() {
 ################################################################################
 
 update_credentials
+upgrade_conf_file
 upgrade_slapdconf
 set_git_branch
 upgrade_apache_conf


rudder-webapp/SOURCES/rudder-upgrade
```diff
@@ -379,6 +380,21 @@ upgrade_masterfiles() {
...
+upgrade_conf_file() {
+
+  STEP="Upgrade /opt/rudder/etc/rudder-web.properties"
+  CONF_FILE="/opt/rudder/etc/rudder-web.properties"
+
+  # - 7.0.0~alpha1 : Rename property rudder.community.port into rudder.policy.distribution.port.cfengine
+  OLD_PROPERTY_NAME="rudder.community.port"
+  NEW_PROPERTY_NAME="rudder.policy.distribution.port.cfengine"
+  ATTRIBUTESET=$(grep -c "^${NEW_PROPERTY_NAME}[ \t]*=" ${CONF_FILE} || true)
+  if [ ${ATTRIBUTESET} -eq 0 ]; then
```
peckpeck, line 392: I think his should be an "if old_property is present" instead

rudder-webapp/SOURCES/rudder-upgrade
```diff
@@ -379,6 +380,21 @@ upgrade_masterfiles() {
...
+    sed -e "s/$PROPERTY_NAME/rudder.policy.distribution.port.cfengine/" ${CONF_FILE} | grep -C 2 "port.cfengine"
```
peckpeck, line 393: I don't understand this line

fanf: That's... The test line, not the one that should be here.

rudder-webapp/SOURCES/rudder-upgrade
fanf: No, if the both are here (for ex, someone has added the new one in place or replacing the old one), then we don't want to do the sed, else we could have a duplicated property (and I'm not sure what happens in that case. First value taken? Last one?)

fanf:
PR updated with a new commit

rudder-webapp/SOURCES/rudder-upgrade
```diff
@@ -379,6 +380,21 @@ upgrade_masterfiles() {
...
+
+  STEP="Upgrade /opt/rudder/etc/rudder-web.properties"
+  CONF_FILE="/opt/rudder/etc/rudder-web.properties"
+
+  # - 7.0.0~alpha1 : Rename property rudder.community.port into rudder.policy.distribution.port.cfengine
+  OLD_PROPERTY_NAME="rudder.community.port"
+  NEW_PROPERTY_NAME="rudder.policy.distribution.port.cfengine"
+  ATTRIBUTESET=$(grep -c "^${NEW_PROPERTY_NAME}[ \t]*=" ${CONF_FILE} || true)
+  if [ ${ATTRIBUTESET} -eq 0 ]; then
+    sed -i -e "s/$PROPERTY_NAME/rudder.policy.distribution.port.cfengine/" ${CONF_FILE}
```
peckpeck, line 393: $PROPERTY_NAME is not defined

peckpeck: if there are 2 entries (an old and a new), now there will be 2 conflciting lines

fanf:
PR updated with a new commit

rudder-webapp/SOURCES/rudder-upgrade
```diff
@@ -379,6 +380,21 @@ upgrade_masterfiles() {
...
+
+  STEP="Upgrade /opt/rudder/etc/rudder-web.properties"
+  CONF_FILE="/opt/rudder/etc/rudder-web.properties"
+
+  # - 7.0.0~alpha1 : Rename property rudder.community.port into rudder.policy.distribution.port.cfengine
+  OLD_PROPERTY_NAME="rudder.community.port"
+  NEW_PROPERTY_NAME="rudder.policy.distribution.port.cfengine"
+  ATTRIBUTESET=$(grep -c "^${NEW_PROPERTY_NAME}[ \t]*=" ${CONF_FILE} || true)
+  if [ ${ATTRIBUTESET} -eq 0 ]; then
+    sed -i -e "s/${OLD_PROPERTY_NAME}/${NEW_PROPERTY_NAME}/" ${CONF_FILE}
```
peckpeck, line 393: this will still leave 2 configuration line if there is one of each at the beginning

fanf: yes, but that's not a problem since we take the new one first

peckpeck: how do you know it's the new one ? they both have the same name after this

fanf: I don't understand ?
If the user added `rudder.policy.distribution.port.cfengine`, then we have both `rudder.policy.distribution.port.cfengine` and `rudder.community.port` but the `if` prevent the sed; rudder takes `rudder.policy.distribution.port.cfengine` as expected. 
If the user didn't do anything (nominal case), then there is only `rudder.community.port`, the sed is done and rudder use the only value. 
If somehow the upgrade script doesn't run, then doesn't find `rudder.policy.distribution.port.cfengine`, so it fall back to `rudder.community.port`.

peckpeck: aah i understood the if the other way around

peckpeck (approved):

fanf:
OK, squash merging this PR

PR #2504, opened 2021-07-30
Fixes #19638: Add /var/rudder/lib/ssl to agent package

amousset:
https://issues.rudder.io/issues/19638


peckpeck:
I think there should be a mkdir in the Makefile

amousset:
PR updated with a new commit

fanf (approved):

PR #2515, opened 2021-08-10
Fixes #19757:  Remove useless log messages from packages scripts - agent

amousset:
https://issues.rudder.io/issues/19757


rudder-webapp/SPECS/rudder-webapp.spec
```diff
@@ -396,7 +396,6 @@ rm -rf %{buildroot}
 %config(noreplace) /opt/rudder/etc/rudder-passwords.conf
 %config(noreplace) /etc/default/rudder-jetty
 %attr(0600, root, root) /opt/rudder/etc/rudder-passwords.conf
-%attr(0755, rudder-slapd, rudder-slapd) /var/rudder/ldap/openldap-data
```
amousset, line 399: we already do the chown in postinst and this one fails with:

```
warning: user rudder-slapd does not exist - using root
warning: group rudder-slapd does not exist - using root
```

ncharles (approved):

PR #2530, opened 2021-10-05
Fixes #20060: techniques distributePolicy and server-roles are not removed from /var/rudder/configuration-repository/techniques/system when upgrading from 6.2 to 7.0

ncharles:
https://issues.rudder.io/issues/20060


ncharles:
PR updated with a new commit

amousset (approved):

PR #2560, opened 2021-12-28
/etc/cron.d/rudder-agent is a conffile in rpm too

peckpeck:
https://issues.rudder.io/issues/20483


rudder-agent/SPECS/rudder-agent.spec
```diff
@@ -573,7 +582,6 @@ rm -f %{_builddir}/file.list.%{name}
 
 %if "%{?aix}" == ""
 # no init no cron and no profile with aix
-%config /etc/cron.d/rudder-agent
```
amousset, line 576: don't we need to add it as normal file?

peckpeck: It is already declared as a regular %file, otherwise %config would produce an error

amousset (approved):

PR #2572, opened 2022-02-01
Fixes #20687: Missing augeas dependency on rpm

amousset:
https://issues.rudder.io/issues/20687

### DIFF
--- a/rudder-agent/SPECS/rudder-agent.spec
+++ b/rudder-agent/SPECS/rudder-agent.spec
@@ -266,10 +266,10 @@ Requires: curl
 %endif
 
 ## Augeas dependencies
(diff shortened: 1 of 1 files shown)


amousset:
PR updated with a new commit

rudder-agent/SPECS/rudder-agent.spec
```diff
@@ -266,10 +266,10 @@ Requires: curl
 %endif
 
 ## Augeas dependencies
-%if "%{use_system_augeas}" == "true"
+%if "%{with_augeas}" == "false"
```
ncharles, line 269: THIS is misleading

ncharles (approved):

PR #2593, opened 2022-05-05
Fixes #21067: Add windows support to the inventory software update section

Fdall:
https://issues.rudder.io/issues/21067

### DIFF
--- a/rudder-agent/SOURCES/patches/fusioninventory/21067-windows-updates.patch
+++ b/rudder-agent/SOURCES/patches/fusioninventory/21067-windows-updates.patch
@@ -1,0 +1,145 @@
+From 389ab6ac2120631768a44eef9827ad1d0753f2d6 Mon Sep 17 00:00:00 2001
+From: =?UTF-8?q?F=C3=A9lix=20Dallidet?= <felix.dallidet@rudder.io>
+Date: Thu, 19 May 2022 19:41:37 +0200
+Subject: [PATCH] Add Windows support to the software update inventory feature
+
+---
+ .../Inventory/Generic/Softwares/Windows.pm    | 126 ++++++++++++++++++
+ 1 file changed, 126 insertions(+)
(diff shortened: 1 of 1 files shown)


rudder-agent/SOURCES/patches/fusioninventory/21067-windows-updates.patch
```diff
@@ -0,0 +1,129 @@
...
++        print $fh $raw_json;
++        print $fh "\n";
++        print $fh $@;
++        close($fh);
++        $logger->error("Invalid JSON data stored in $filename") if $logger;
++    }
++    foreach my $update (@updates) {
++      my $package = {
++        NAME    => $update->{Name},
++        ARCH    => "x86_64",
```
amousset, line 97: this needs to be replaced by the real value

rudder-agent/SOURCES/patches/fusioninventory/21067-windows-updates.patch
```diff
@@ -0,0 +1,129 @@
...
++        VERSION => "1.0",
```
amousset, line 98: if we don't have a version number we need to make it optional on the webapp side

rudder-agent/SOURCES/patches/fusioninventory/21067-windows-updates.patch
```diff
@@ -0,0 +1,129 @@
...
++        open(my $fh, '>', $filename);
++        print $fh $raw_json;
++        print $fh "\n";
++        print $fh $@;
++        close($fh);
++        $logger->error("Invalid JSON data stored in $filename") if $logger;
++    }
++    foreach my $update (@updates) {
++      my $package = {
++        NAME    => $update->{Name},
```
amousset, line 96: I think we should use the KB id here to match what is done for installed packages.

We could reuse the string concatenated with the KB description in the DESCRIPTION field.

Fdall:
Commit modified

PR #2600, opened 2022-05-17
Fixes #21111: Remove code for handling python2 in server packages

peckpeck:
https://issues.rudder.io/issues/21111

### DIFF
--- a/rudder-api-client/SPECS/dependencies
+++ b/rudder-api-client/SPECS/dependencies
@@ -1,4 +1,2 @@
-sles-12:python python-requests
 sles-15:python3 python3-requests python3-docopt python3-urllib3
-rhel-7:python python-requests python-urllib3
 rhel-8:python3 python3-requests python3-urllib3
--- a/rudder-relay/SPECS/rudder-relay.spec
+++ b/rudder-relay/SPECS/rudder-relay.spec
@@ -70,15 +70,8 @@ Requires: %(../format-dependencies rpm %{old_epoch}:%{real_version} rudder-agent
 
 ## RHEL
 %if 0%{?rhel}
-Requires: mod_ssl shadow-utils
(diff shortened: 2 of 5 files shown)


rudder-api-client/SPECS/dependencies
```diff
@@ -1,4 +1,2 @@
-sles-12:python python-requests
 sles-15:python3 python3-requests python3-docopt python3-urllib3
-rhel-7:python python-requests python-urllib3
```
amousset, line 3: dont't we need to add the python3 version to hrhel7 too?

rudder-relay/SPECS/rudder-relay.spec
```diff
@@ -70,62 +70,25 @@ Requires: %(../format-dependencies rpm %{old_epoch}:%{real_version} rudder-agent
...
 run a Rudder relay server on a machine.
 
-#=================================================
-# Source preparation
-#=================================================
-%prep
-%setup -c
-
-# We don't know the exact version
-cd rudder-sources-*/rudder/relay/sources/
```
amousset, line 122: why removing this?

amousset: Indeed, but I think the setup -c is still required to correctly uncompress the source

rudder-api-client/SPECS/dependencies
amousset: but don't we keep it?

rudder-relay/SPECS/rudder-relay.spec
peckpeck: i thought it was the default

amousset: the -c is not

amousset:
won't be done in 7.2, should be done for 7.3, closing for now

PR #2603, opened 2022-05-18
Fixes #21120: rudder-relay postinst fails to restart apache

peckpeck:
https://issues.rudder.io/issues/21120


amousset:
What does this fix? The issue is empty.

PR #2606, opened 2022-05-23
Fixes #21151: Make rudder-jetty.conf and default rudder-jetty compatible with JVM 11 and 17

fanf:
https://issues.rudder.io/issues/21151

Main changes: 
- use G1GC by default 
  - with string deduplication
  - and touch pages at launch
- add high level, not verbose GC info (always) in /var/log/rudder/webapp/jvm-gc.log (rotation: 5 files, 50M each)
- let user be able to override GC log level (with a suitable example for debug metrics in comments)
- notice in comment of the overridding file what is used by default

### DIFF
--- a/rudder-server/SOURCES/rudder-jetty.default
+++ b/rudder-server/SOURCES/rudder-jetty.default
@@ -23,19 +23,38 @@ JAVA_MAXPERMSIZE=256
 #
 #JAVA_OPTIONS=""
 
-# Java Garbage collector option
-# On large heap (6 Go or more) and loaded platforms, ConcurrentMarkSweep (default GC 
-# for Rudder) becomes inefficient and Rudder can experience long "stop the world" 
-# garbage collections.
-# To avoid that, you can use G1GC which leads to faster (or no) full GC on big heap,
-# at the prize of a less throughout and a need of some more memory to work 
-# efficiently (~10%).
-# You can also tune you GC option here.
+# Java garbage collector option
 #
+# By default, Rudder uses G1GC which is a very good general purpose GC on
+# all sizes of heap. The configuration should be good on most case.
+# But you can change GC or tune GC options here, like debug flags.
+# Documentation can be found here:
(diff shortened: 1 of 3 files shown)


rudder-server/SOURCES/rudder-jetty.default
```diff
@@ -23,19 +23,26 @@ JAVA_MAXPERMSIZE=256
...
+#
+# By default, Rudder uses G1GC which is a very good general purpose GC on
+# all sizes of heap. The configuration should be good on most case.
+# But you can change GC or tune GC options here, like debug flags.
+# Documentation can be found here:
+# https://docs.oracle.com/en/java/javase/11/gctuning/garbage-first-garbage-collector-tuning.html
 #
 #JAVA_GC="-XX:+UseG1GC
-#-XX:+UnlockExperimentalVMOptions
-#-XX:+UseStringDeduplication
```
amousset, line 37: we should probably keep our default conf as basis here

fanf: I don' understand ? Do you want that we keep the `rudder-jetty.default` as it was in previous versions ?

ncharles: default of rudder-jetty.conf should be reflected in this file, so that we know what's the default config; and offer as an extra others configuration

peckpeck:
You should make ruder-jetty.conf not a conffile and test an upgrade from the previous version with a modified conf to make sure the file is replaced both with RPM and DEB

fanf:
PR updated with a new commit

rudder-server/debian/conffiles
```diff
@@ -1,4 +1,4 @@
-/opt/rudder/etc/rudder-jetty.conf
+/etc/default/rudder-jetty
```
amousset, line 1: no need, everything in /etc is automatically flagged as conffile

fanf:
PR updated with a new commit

amousset (approved):

PR #2641, opened 2022-07-07
Fixes #21382: yum installs external-db instead of postgresql by default -   bis

peckpeck:
https://issues.rudder.io/issues/21382


peckpeck:
PR updated with a new commit

amousset (approved):

PR #2652, opened 2022-08-09
Fixes #21583: Apache conf files are not overwritten after update

amousset:
https://issues.rudder.io/issues/21583

### DIFF
--- a/rudder-server/SPECS/rudder-server.spec
+++ b/rudder-server/SPECS/rudder-server.spec
@@ -234,6 +234,14 @@ if [ $1 -ne 1 ]
   fi
 
   getfacl --absolute-names --recursive /opt/rudder/etc/hooks.d/ > /tmp/rudder-hooks-upgrade
+
+  # Backup in case there was custom changes when it was a conffile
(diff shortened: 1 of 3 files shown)


rudder-server/SPECS/rudder-server.spec
```diff
@@ -234,6 +234,14 @@ if [ $1 -ne 1 ]
   fi
 
   getfacl --absolute-names --recursive /opt/rudder/etc/hooks.d/ > /tmp/rudder-hooks-upgrade
+
+  # Backup in case there was custom changes when it was a conffile
+  # not used anymore
```
peckpeck, line 239: the equivalent of `case upgrade)` for rpm is` if [ $1 -ne 1 ]`

amousset: it's already in a big `if [ $1 -ne 1 ]`

peckpeck: aah sory

peckpeck (approved):

PR #2657, opened 2022-08-25
Fixes #21640: Uprading relay to 7.2 fails on ubuntu 22

peckpeck:
https://issues.rudder.io/issues/21640

### DIFF
--- a/rudder-relay/debian/postinst
+++ b/rudder-relay/debian/postinst
@@ -51,6 +51,14 @@ case "$1" in
         mv /opt/rudder/etc/rudder-pkg/rudder-pkg.conf.distrib /opt/rudder/etc/rudder-pkg/rudder-pkg.conf
       fi
 
+      # sometimes (ubuntu22 ?) divert prevents the file to be present
+      if [ ! -f /opt/rudder/etc/rudder-networks-24.conf.distrib ]; then
+        mv /opt/rudder/etc/rudder-networks-24.conf.distrib /opt/rudder/etc/rudder-networks-24.conf
+      fi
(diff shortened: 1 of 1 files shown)


amousset (approved):

rudder-relay/debian/postinst
```diff
@@ -51,6 +51,11 @@ case "$1" in
         mv /opt/rudder/etc/rudder-pkg/rudder-pkg.conf.distrib /opt/rudder/etc/rudder-pkg/rudder-pkg.conf
       fi
 
+      # sometimes (ubuntu22 ?) divert prevents the file to be present
+      if [ ! -f /opt/rudder/etc/rudder-networks-24.conf ]; then
+        touch /opt/rudder/etc/rudder-networks-24.conf
```
amousset, line 56: shouldn't we move the .distrib file?

peckpeck: we shouldn't need to, but we can, this way the directory would be cleaned from .distrib files

peckpeck:
PR updated with a new commit

rudder-relay/debian/postinst
```diff
@@ -51,6 +51,14 @@ case "$1" in
         mv /opt/rudder/etc/rudder-pkg/rudder-pkg.conf.distrib /opt/rudder/etc/rudder-pkg/rudder-pkg.conf
       fi
 
+      # sometimes (ubuntu22 ?) divert prevents the file to be present
+      if [ ! -f /opt/rudder/etc/rudder-networks-24.conf.distrib ]; then
```
amousset, line 55:
```suggestion
      if [ -f /opt/rudder/etc/rudder-networks-24.conf.distrib ]; then
```

PR #2706, opened 2023-02-23
Fixes #22377: Rudder-jetty timeout is the systemd default one and may be short

peckpeck:
https://issues.rudder.io/issues/22377


rudder-server/SOURCES/rudder-jetty.service
```diff
@@ -13,6 +13,7 @@ LimitNOFILE=64000
 ExecStart=/opt/rudder/bin/rudder-jetty.sh start
 ExecStop=/opt/rudder/bin/rudder-jetty.sh stop
 SuccessExitStatus=143
+TimeoutStartSec=120s
```
amousset, line 16: @ncharles doesn't this also requires a change on the jetty side?

peckpeck:
@amousset @ncharles ping

rudder-server/SOURCES/rudder-jetty.service
ncharles: yes, the taimeout within jetty needs to be increased as well - both apply :(

ncharles: Like adding a default value of JETTY_START_TIMEOUT=120 in /etc/default/rudder-jetty

peckpeck:
PR updated with a new commit

amousset (approved):

PR #2766, opened 2023-06-15
Fixes #22899: there is a hidden dependency on syslog service

peckpeck:
https://issues.rudder.io/issues/22899


peckpeck:
PR updated with a new commit

amousset (approved):

PR #2784, opened 2023-07-18
Fixes #23083: rhel7 server build should depends on python3-pip and python3

VinceMacBuche:
https://issues.rudder.io/issues/23083


VinceMacBuche:
Commit modified

PR #2786, opened 2023-07-21
Fixes #23116: parent patch doesn't apply on 8.0

peckpeck:
https://issues.rudder.io/issues/23116


VinceMacBuche:
I already reverted the wrong patch (it was in a separated commit)

PR #2792, opened 2023-07-31
Fixes #23166: Installation of Rudder 8.0-nightly fails on AlmaLinux 9

ncharles:
https://issues.rudder.io/issues/23166


ncharles:
wrong ticket number

PR #2794, opened 2023-08-01
Fixes #23186: Rudder Server 7.3.4 doesn't install on SLES 15 SP4

ncharles:
https://issues.rudder.io/issues/23186

On SLES15, the fresh installation fails, because postgresql is not started by default
The change cannot simply be "start postgresql" because:
* On RHEL like, we need to initdb
* with a remote database, we don't need a local postgres
* the script is a total mess

So i massively rewrote it to:
* detect if database is local or not
* have 2 distinct paths for local and remote database, to untangle everything
* try to not reinit database at each upgrade


ncharles:
This is not tested yet

PR #2807, opened 2023-09-06
Fixes #23388: Server install fails with postgresql encoding error

fanf:
https://issues.rudder.io/issues/23388

It looks like template1, the default template for creating database, has index and encoding can't be changed. So we need to ressort on `template0`, which is pristine. 
See: https://stackoverflow.com/a/28131347
And for the syntax: https://www.postgresql.org/docs/current/manage-ag-templatedbs.html


amousset:
OK, merging this PR

fanf:
OK, merging this PR

PR #2839, opened 2023-11-18
Fixes #23710: Update to jetty 10.0.18

amousset:
https://issues.rudder.io/issues/23710

Quite a big change as expected. Getting rid of all the patches that fell into one of these categories:

* service start/stop fixes. This has been rewritten upstream since our patches plus we switched to systemd for service management. Let's start from a clean sheet.
* OS specific fixes now outdated: softkn version for CentOS 7, ulimit for SLES 11

Kept two patches, but switched to systemd unit config instead for better maintainability:

* umask
* memory check

Regarding the systemd unit:

* The best option would be to use the `notify` type, along with something like https://github.com/Normation/rudder/pull/5180. It allows to provide exact information about the service process to systemd, and let it handle error handling (instead of what we do in the agent for example). This is also used by jenkins for example.
  * But it does not work in our case as the script exists after starting the webapp, while systemd expects the main process to keep running, so it fails even in the notification works correctly.
  * Once we have a single jar file for the whole application, we will be able to switch to it.
* Don't add `PIDFile`: it causes problems as the main process, pointed to by the pid file, is not a direct children of systemd (there is an `xargs` process in between. Let systemd autodetect the main process.

The memory check appears like before in the logs (both are displayed in the screenshot):

image

This will requires basic test on all server OSes. The problems it might cause are mostly related to start/stop and should be easy to workaround. We could backport to 7.3 also if it proves itself to work smoothly.


amousset:
PR updated with a new commit

fanf (approved):
Seems ok, will test with nightly

PR #2846, opened 2023-12-22
Fixes #23958: Abort server build if rudder.war is less than 1Mb

VinceMacBuche:
https://issues.rudder.io/issues/23958

### DIFF
--- a/rudder-server/SOURCES/mvn-like
+++ b/rudder-server/SOURCES/mvn-like
@@ -28,3 +28,10 @@ else
 fi
 curl -s -o "${outfile}" "${url}"
 
+# Set a minimum size of 1Mb
+minimumsize=1000000
+actualsize=$(wc -c <"${outfile}")
+if [ ${actualsize} -le ${minimumsize} ]; then
+    echo "${outfile} size is less than 1Mb, abort build"
+    exit 1
+fi


VinceMacBuche:
Commit modified

rudder-server/SOURCES/mvn-like
```diff
@@ -28,3 +28,10 @@ else
 fi
 curl -s -o "${outfile}" "${url}"
 
+# Set a minimum size of 1Mb
+minimumsize=1000000
+actualsize=$(wc -c <"${outfile}")
+if [ ${actualsize} -le ${minimumsize} ]; then
```
Fdall, line 34: why not use `stat -c %s `, `wc` does not seem evident to me

VinceMacBuche: wc -c is just counting the bytes, it's in the usual answers about checking size of a file.

Fdall (approved):

rudder-server/SOURCES/mvn-like
```diff
@@ -28,3 +28,10 @@ else
 fi
 curl -s -o "${outfile}" "${url}"
 
+# Set a minimum size of 1Mb
+minimumsize=1000000
```
clarktsiory, line 32: 1Mb is strictly 1024*1024 = 1048576, but I guess both values are close so it's fine ^^

PR #2860, opened 2024-03-14
Fixes #24449: Move plugins postupgrade after webapp restart

amousset:
https://issues.rudder.io/issues/24449

Not sure if it will fix openscap upgrade problems (it could if the webapp was left in a broken state by previous upgrade steps) but at least I think it makes more sense.


amousset:
OK, merging this PR

PR #2907, opened 2024-09-23
Fixes #25496: Inventory fails because rudder-agent has no version in inventory

peckpeck:
https://issues.rudder.io/issues/25496


peckpeck:
For an unknown reason, windows RegOpenKeyExW returns error 2 (file absent) instead of opening the registry key on a random basis (around 1 every 4 times).
This has been reproduced even after removing : threads, locks, other kind of registry calls, closing all opens sequentialy, sleeping between calls. 
So the only solution found is to try the repeatedly as long as it returns error 2.

amousset (approved):

peckpeck:
This has already been done in getNewRegistrySubKeys, but it must be done for all RegOpenKeyExW. This is what this PR is for.

PR #2937, opened 2024-12-17
Fixes #26084: Add nettle as a rudder-package dependency

amousset:
https://issues.rudder.io/issues/26084


amousset:
PR updated with a new commit

PR #2939, opened 2024-12-31
Fixes #26090: We should not be able to install package on wrong OS

peckpeck:
https://issues.rudder.io/issues/26090


peckpeck:
PR updated with a new commit

amousset (approved):

PR #2984, opened 2025-04-25
Fixes #26809: Agent fails to build without embedded augeas

peckpeck:
https://issues.rudder.io/issues/26809


amousset:
Too many commits :)

PR #2993, opened 2025-05-27
Fixes #26966: "Already on 'master'" in postinst logs

amousset:
https://issues.rudder.io/issues/26966


amousset:
PR updated with a new commit

PR #2996, opened 2025-06-03
Fixes #27025: RHEL 10 Rudder server should depend on java 21

VinceMacBuche:
https://issues.rudder.io/issues/27025

### DIFF
--- a/rudder-server/SPECS/rudder-server.spec
+++ b/rudder-server/SPECS/rudder-server.spec
@@ -100,9 +100,15 @@ BuildRequires: gcc, rsync
 BuildRequires: groff-base
 BuildRequires: selinux-policy-devel
 Requires: mod_ssl httpd shadow-utils
-Requires: jre-17-headless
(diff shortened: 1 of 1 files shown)


rudder-server/SPECS/rudder-server.spec
```diff
@@ -100,11 +100,17 @@ BuildRequires: gcc, rsync
 BuildRequires: groff-base
 BuildRequires: selinux-policy-devel
 Requires: mod_ssl httpd shadow-utils
-Requires: jre-17-headless
 Requires: perl-Digest-SHA libtool-ltdl
 BuildRequires: openssl-devel libtool-ltdl-devel git
 %endif
 
+%if 0%{?rhel} && 0%{?rhel} <= 9 || 0%{?fedora}
```
amousset, line 107: fedora is closer to rhel10, but we should probably just remove it

VinceMacBuche:
Commit modified

VinceMacBuche:
Commit modified

PR #3022, opened 2025-08-21
Fixes #27413: Upload large file in technique resources or file download returns error

clarktsiory:
https://issues.rudder.io/issues/27413

https://httpd.apache.org/docs/2.4/en/mod/mod_proxy_http.html#env : I tried the different kinds of env variables that could explain the interrupted upload, and `proxy-nokeepalive` seems to be the only one fixing it

> Forces the proxy to close the backend connection after each request.

:arrow_right:  that explains why Jetty terminates but apache does not seem to understand why (e.g. when uploading a >8MB file, see https://github.com/Normation/rudder/pull/5828/files#r1728452973)


clarktsiory:
Commit modified

PR #3143, opened 2026-07-07
Fixes #29197: Relax perl version requirement to minor version

peckpeck:
https://issues.rudder.io/issues/29197


amousset:
It only changes RHEL where we don't have a problem. The problem is afaik only on SLES 16,

peckpeck:
oops indeed

peckpeck:
PR updated with a new commit

amousset (approved):

PR #3144, opened 2026-07-07
Fixes #29212: perl dependency should not be equal

peckpeck:
https://issues.rudder.io/issues/29212


amousset (approved):

peckpeck:
PR updated with a new commit

PR #3172, opened 2026-08-04
Fixes #29455: Generate root policies in the standard dir - packages

amousset:
https://issues.rudder.io/issues/29455

To merge after https://github.com/Normation/rudder/pull/7362


clarktsiory (commented):
I don't know about augeas, @m4rtinh4rt 
but the change looks consistent with https://github.com/Normation/rudder/pull/7362

PR #3182, opened 2026-08-19
Fixes #29575: webapp non-root: fix permissions on hooks

m4rtinh4rt:
https://issues.rudder.io/issues/29575

### DIFF
--- a/rudder-server/SOURCES/rudder-server-postinst
+++ b/rudder-server/SOURCES/rudder-server-postinst
@@ -283,14 +283,17 @@ fi
 
 chown -R rudder-webapp:rudder \
   /var/rudder/run \
-  /var/log/rudder/webapp \
-  /opt/rudder/etc/rudder-jetty-base/ \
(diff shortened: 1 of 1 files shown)


rudder-server/SOURCES/rudder-server-postinst
```diff
@@ -291,6 +291,7 @@ chown -R rudder-webapp:rudder \
   /opt/rudder/share/webapps/ \
   /opt/rudder/etc/webapp \
   /var/rudder/lib/ssl/shared \
+  /opt/rudder/etc/hooks.d
```
amousset, line 296: i think would should separate rudder-webapp:rudder and rudder-webapp:rudder-webapp/root as giving the rudder groups gives access to rudder-relayd to things it does not need.
Here it would be a root injection capability.

m4rtinh4rt:
PR updated with a new commit

amousset (approved):

PR #3183, opened 2026-08-24
Fixes #29611: webapp non-root: filter out the old hooks from the ACL list

m4rtinh4rt:
https://issues.rudder.io/issues/29611


m4rtinh4rt:
PR updated with a new commit

amousset (approved):

PR #3194, opened 2026-09-16
Fixes #29441: Installation logs from Rudder don't show the version

m4rtinh4rt:
https://issues.rudder.io/issues/29441


m4rtinh4rt:
PR updated with a new commit

m4rtinh4rt:
PR updated with a new commit

rudder-server/debian/postinst
```diff
@@ -26,7 +26,8 @@ case "$1" in
         RUDDER_FIRST_INSTALL="true"
       fi
       LOG_FILE="/var/log/rudder/install/rudder-server-$(date +%Y%m%d).log"
-      echo "`date` - Starting rudder-server post installation script" >> ${LOG_FILE}
+      PACKAGE_VERSION="$(dpkg-query -W -f='${Version}' rudder-server)"
```
amousset: it is generally a bad idea to call dpkg from inside a dpkg run

amousset: have a look at the arguments this script takes in deb doc

rudder-server/SPECS/rudder-server.spec
```diff
@@ -167,6 +167,8 @@ make -f /usr/share/selinux/devel/Makefile
 #=================================================
 %install
 
+echo "`date` - Starting %{name} %{real_version} installation" >> ${LOG_FILE}
```
amousset: install is not the right place, it run during package build, not on the user's systems

PR #3197, opened 2026-09-18
Fixes #29772: webapp non-root: fix postinst warning about failure to change directory

m4rtinh4rt:
https://issues.rudder.io/issues/29772


amousset:
What is the error? It should be in the ticket description at least. And why will this fix the pb?

m4rtinh4rt:
When using the `su` command, the rudder-webapp user cannot cd into /root. The chdir operation fails and produces an error message.

There is no functional impact; it is simply an annoying warning.
Changing the working directory to /tmp before running `su` resolves the issue, since /tmp is accessible to all users.

I have updated the issue.

